PR #13, opened 2016-11-29
Create email validator

hootlex:
Added a simple email validator with tests.


test/unit/specs/validators/email.spec.js
```diff
@@ -0,0 +1,48 @@
+import email from 'src/validators/email'
+
+describe('email validator', () => {
+  it('should not validate empty string', () => {
+    expect(email('')).to.be.false
```
Frizi: This actually should be true, otherwise you imply that the field is required. We have a separate validator for that.

skyrpex:
I don't think that regex is okay: it only matches emails from domains that have between 2 and 6 characters...

test/unit/specs/validators/email.spec.js
```diff
@@ -0,0 +1,48 @@
...
+    expect(email('someone@gmail')).to.be.false
+    expect(email('someone@gmail.')).to.be.false
+    expect(email('someone@gmail.c')).to.be.false
+    expect(email('gmail.com')).to.be.false
+    expect(email('@gmail.com')).to.be.false
+  })
+
+  it('should not validate addresses that contain unsupported characters', () => {
+    expect(email('someone@g~mail.com')).to.be.false
+    expect(email('some!one@gmail.com')).to.be.false
```
Frizi: Exclamation mark is perfectly valid character for email address.

Validating emails in general is quite hard, as RFC has many details that people often forget about. Consider making your regex a bit more relaxed. It's always better to allow some false positives than to be too strict in validation. Proper email validation would involve querying the DNS for MX record, and this requires a backend service, so we can't really expect our validator to be perfect anyway.

furey: At work we tend to settle on something like the following "casual" approach…

`[characters, no spaces]@[characters, no spaces].[characters, no spaces]`

…which helps prevent typos, which generally suits our immediate front-end needs.

For complete validation, as @Frizi suggested, after "casual" validation we'd ping a backend validation API (and occasionally follow up with a confirmation email, depending on business requirements).

Cheers. 🙂

hootlex: Thanks for the feedback guys. I tried to make the regex as simple as possible.
What about this [one](http://emailregex.com/)?

Frizi:
I looked over the regex, but without slicing it to tiniest pieces I'm not able to grasp it. Kind of hard to check if it's correct. Could you document it a bit and leave any info how it was constructed? Also please add the entry to docs about this validator. Thanks!

furey:
Over time I've discovered that validating emails via regex is a precarious endeavour.

Unless your regex is 100% dependable – and there seems to be numerous Google results to argue otherwise – I've surrendered to the fact that you should treat whatever regex you implement as "email-ish", consider implementing something like https://github.com/mailcheck/mailcheck to minimise user error, and if your business requirements allow, follow up with a confirmation email.

Food for thought:

https://davidcel.is/posts/stop-validating-email-addresses-with-regex/

Be keen to get other people's take on the matter.

Thanks for listening!

hootlex:
@Frizi @furey after your first feedback I searched for the 'best' email regex and I picked the one found at  [emailregex.com](http://emailregex.com/), that's why it's so huge and hard to grasp. I think it suits the needs of the validator.

mubaidr:
I think it does not matter if this is 100% accurate or not, but since this is one of the most basic validation required in almost every application, it should be available in default validations. All caveats related to this can be explained in documentation.

Frizi:
@hootlex Thanks! I will accept that now, as there is no way to be perfectly correct anyway. I still think this regex is a bit too complicated, but i was not able to find a counterexample except something like user@localhost or IP addresses, but those ones should probably not be accepted on public webapps anyway.

In case of any problems, please file an issue.

PR #42, opened 2017-01-07
fix typo

lukeb:


shentao:
Thanks! Will be updated after next docs deploy!

PR #45, opened 2017-01-09
Fix typo in example partial

johnnynotsolucky:


Frizi:
Hi! Well spotted, thank you for the PR. Will be deployed with next patch :)

PR #47, opened 2017-01-18
add typescript definitions

mrellipse:
i thought these might be useful to others writing vue apps using typescript.

happy to maintain them going forwards, if you do decide to include these definitions and the api changes

i manually tested simple validation, nested validation, and collections using $each

sample typescript implementation is below

*******
import Vue = require('vue');
import { ValidationRules, Vuelidate, IVuelidate } from 'vuelidate';
import { required, minLength } from 'vuelidate/lib/validators';

interface IViewModel {
    message: string;
}

let data: IViewModel = {
    message: ''
};

let rules: ValidationRules<IViewModel> = {
    message: {
        required,
        minLength: minLength(5)
    }
};

let template = `<div>
  <div v-if="$v.message.$dirty">
    <span class="alert-danger" v-if="!$v.message.required">required</span>
    <span class="alert-danger" v-if="!$v.message.minLength">min 10 chars</span>
  </div>
  <div>
    <input type='input' v-model.trim='message' @input="$v.message.$touch()">
    <pre>{{$v.message}}</pre>
  </div>
</div>`;

export default {
    template: template,
    data: function () {
        return data;
    },
    validations: rules
} as Vue.ComponentOptions<IVuelidate<IViewModel>>

*******


Frizi:
Hi,

Thats Awesome! I'd like to merge that for sure. One thing is I won't be able to maintain it too well, as I'm not using typescript day to day.

One thing:
```
$each?: {
  [key: number]: IValidator;
}
```

They $each key can be any string or number really. You can as well use it for iterating over unknown objects.

mrellipse:
I'd be happy to help maintain typescript for the project going forwards, no problem :)

I just submitted a fix for $each, so you can now use either index[] or object.property notation to access validation results.

In terms of tests ... I had to think for a bit. Since we are talking about testing changes to the shapes/types that describe the library,  I reckon a 'can this compile?' test would be appropriate.

I will look into writing up some classes that consume the types exported from .d.ts, and replicate the example scenarios outlined in [docs](https://monterail.github.io/vuelidate/#examples).

Cheers!

Frizi:
Thanks for your support :) I'm not merging this yet only because i'm not able to verify that without significant time investments. Looking forward for the news regarding tests, as it should change the situation.

mrellipse:
Hi Frizi,

All right, I learnt a few new useful things along the way, so it was time well spent! I have something more workable for you to review now. FYI - I squashed the commits into one so it doesn't clutter up history..

I checked the official vuejs repo and confirmed they are doing 'compile-only' tests of TS interfaces being shipped - that was in line with my initial thoughts on how to handle the situation. But I went ahead and created a typescript-based test suite you can run in Karma, which covers common use cases

- simple properties
- nested properties
- iterating over collections (using $each)
- nested groups

Could you please review and let me know what you think? I will probably need some help create/updating docs though - learning pug, and prism with typescript etc might be a bridge too far :)

Talk soon!

Cheers

Frizi:
Hi!

This surely was time well spent, love learning new stuff along the way 😄 I also have a lot to learn in this area, so will probably share the experience.

I'm a bit swamped this weekend and the changeset is pretty big, but I do have it pretty high on my todo list now, so will get back to it soon. Very much appreciated, thanks again!

mrellipse:
Oh yep, learning all the time. Then forgetting. Goes with the territory!

Just noticed the coverage decrease message above - I'll rebase against the latest version of master, and that should make codecov happier.

You did a really nice job with the docs and test coverage.

mrellipse:
Better :)

Frizi:
@mrellipse I'm trying to understand what happens here. I see the definitions file, which seem to work fine, but what's the purpose of doubled tests? API is still changing, new validators are created, it's unnecessary to have the tests double for them. What i was expecting was to see the tests for type checker, but maybe that's simply impossible in TS without rewriting the actual code first. Can you cast some more light on the reasoning behind spec.ts files?

Frizi:
@mrellipse Please take a look at my previous comment. I'd really like to support typescript as much as possible, but having not much knowledge in that area and seeing things with no obvious solution I really need some explanation here before committing to support that.

mrellipse:
Apologies for the delay! I have been pretty busy lately, but hadn't forgotten about this.

The only relevant tests are the integration tests in /ts/specs/_component.spec.ts_ file. They describe use cases from the online docs for vuelidate (nesting, validation groups, collection validations etc.)

If _component.spec.ts_ fails to compile, it means breaking changes have been made to the definitions file.

If _component.spec.ts_ compiles okay, but one of the tests fails, then it means the actual tests themselves need refactoring (the behaviour of the underlying javascript library must have changed in some way.)

 A lot of repos that publish definition files with their js libraries use a 'compile-only' test as proof that the definitions file they are shipping is valid. That didn't seem good enough to me though! You need to have integration tests which test the actual behaviour at runtime, using the published definitions.

Does above make sense? I am having a hard time explaining, but in a nutshell _component.spec.ts_ that is the guts of it.

Screenshot below shows Intellisense in Visual Code using the definitions file.

vuelidate

Let me know what you would like to see as a 'final product', and I will try and get it there.

We can definitely get rid of the doubled tests in /ts/specs/validation.spec.ts & /ts/specs/validators/*.spec.ts. They are unnecessary and would just add extra maintenance.

Frizi:
Hi, sorry for not looking into it, I have to review it all at some point but never really had enough time and motivation to deep dive into TS now. I'm still interested in adding that to the project though. It just have to wait for the right moment unfortunately.

Deadpoolddt:
HI @Frizi 1.5 year ago since, now its right moment? :) 
this presented typings look pretty good btw, also Iam sure plenty of us are willing to contribute if needed, just let us know!

PR #59, opened 2017-02-03
Add numeric validator

38elements:
This adds numeric validator.


Frizi:
Looks good, thank you!

PR #71, opened 2017-02-12
Fix alpha, aplhaNum and numeric validators for undefined and null values

ecmel:
required validator should be used for undefined, null and empty values.


Frizi:
Thanks for that 👍

PR #72, opened 2017-02-12
Update tests for undefined, null and empty values

ecmel:


test/unit/specs/validators/email.spec.js
```diff
@@ -1,6 +1,14 @@
 import email from 'src/validators/email'
 
 describe('email validator', () => {
+  it('should validate undefined string', () => {
+    expect(email(undefined)).to.be.true
+  })
+
+  it('should validate null string', () => {
```
38elements: I think that `should validate null` rather than `should validate null string` is appropriate.

test/unit/specs/validators/alpha.spec.js
```diff
@@ -1,6 +1,10 @@
 import alpha from 'src/validators/alpha'
 
 describe('alpha validator', () => {
+  it('should validate undefined string', () => {
```
38elements: I think that `should validate undefined` rather than `should validate undefined string` is appropriate.

test/unit/specs/validators/sameAs.spec.js
```diff
@@ -9,6 +9,9 @@ describe('sameAs validator', () => {
   it('should not validate different string', () => {
     expect(sameAs('first')('world', parentVm)).to.be.false
     expect(sameAs('second')('hello', parentVm)).to.be.false
+    expect(sameAs('first')(undefined, parentVm)).to.be.false
+    expect(sameAs('first')(null, parentVm)).to.be.false
+    expect(sameAs('first')('', parentVm)).to.be.false
```
38elements: Since `null` and` undefined` are not `string`, I think it is appropriate to test with different items.

38elements:
I think it is better to add  a test of `undefined` in `required.spec.js`.

test/unit/specs/validators/sameAs.spec.js
```diff
@@ -9,6 +9,9 @@ describe('sameAs validator', () => {
   it('should not validate different string', () => {
     expect(sameAs('first')('world', parentVm)).to.be.false
     expect(sameAs('second')('hello', parentVm)).to.be.false
+    expect(sameAs('first')(undefined, parentVm)).to.be.false
+    expect(sameAs('first')(null, parentVm)).to.be.false
```
38elements: Since `null` and `undefined` are not `string`, I think it is appropriate to test with different items.

ecmel: I just want to be sure that the sameAs validator code handles undefined and null values without any runtime errors.

38elements: Thank you for responsing.
I think that `should not validate different value` rather than should `should not validate different string` is appropriate.

ecmel: Yes, good catch :)

ecmel: I changed the wording for null and undefined.

ecmel:
Hi, is there anything wrong with the commit?

If there is I appreciate your feedback.

Frizi:
No, it's all right. Thanks!

PR #76, opened 2017-02-17
add function type `validations`

superwf:
when I use vuelidate for custom component,
 the validations for each input is dynamic,
I would like to transfer validator by component prop
```
<template>
<form>
  <my-input v-for="input of inputs" :model="input.value" :name="input.name" :validator="input.validator" />
</form>
</template>

import { email } from 'vuelidate/lib/validators'
export default {
  data() {
    return {
      inputs: [{
        value: 'name',
        name: 'name',
        validator: {
          email
        }
      }]
    }
  }
}
```

and in my-input component
```
<template>
<input v-model.trim="value" :name="name" class="my-custom-class" />
</template>

export default {
  props: ['model', 'name', 'validator'],
  beforeMount () {
    this.value = this.model
  },
  data () {
    return {
      value: ''
    }
  },
  // here is the point
  validations (self) {
    return {
      value: self.$options.propsData.validator
    }
  }
}
```


Frizi:
Hi!

Fun fact, I have the same functionality already on my $params related branch, but I still have a few concerns.

When reactive config gets recomputed, all validations lose $dirty state. Its not easily avoidable now, some kind of tree traversal diffing is necessary to fix that. I'm still thinking about reusing the VDOM diffing capabilities of Vue for that, but it's tricky. Maybe it's easy and small enough to implement it directly.

We need a documentation that explains what happens here.

Also, specific to your code, you decided to pass vm as first argument. That's ok, but it also should be on the context to follow the existing convention. `options.validations.call(this, this)`

38elements:
IMHO, I think that `options.validations.call(this)` is better.

superwf:
oh, I will read the $params branch

Frizi:
@superwf this change is now in master and in 0.3.0

superwf:
I already use the params branch for a while. :tada: 0.3.0 published.
I read the source, when validator transfered by props, it is realy difficult to dynamically update validate function and $v state when parent validator prop change.

Frizi:
What do you mean by hard?

superwf:
I made a mistake, the validate works when parent validator changed, I forgot using `nextTick`, the $invalid works, but $dirty become false, as you mentioned.

Issue #95, opened 2017-03-12
$v.$touch() does not cascade properly to nested array elements


felix-d:
Thanks for this fantastic pragmatic library, love it.

I'm trying to make all the fields of my form dirty when pressing a button, but it seems `$v.$touch` does not cascade to nested array elements.

See [this JS Fiddle](https://jsfiddle.net/z42kg1jx/3/) to see what I mean. There are 2 fields here, text and a nested comment. Both are required, but when clicking the button, only `text` gets dirty.

It also seems to be the case for other functions such as `$reset`.

Am I missing something or is this a bug?

amirrustam:
I'm noticing something similar after 0.2.0 to 0.3.0 update. `$touch` does not seem to be recursively setting the `$dirty` flag of validation groups either. Seem to be a regression.

felix-d:
Glad to know I'm not crazy :) 

I'm trying to dig in the code to see what might happen but validators mapping looks somewhat complex. I don't see any test case covering this scenario either :(

amirrustam:
.....Or you could downgrade to 0.2.0 😜

I'm short on time to push a fix, though I think I did find the test case for it: https://github.com/monterail/vuelidate/blob/master/test/unit/specs/validation.spec.js#L285

Frizi:
Hi, thanks for the fiddle! I'm taking a look at this.

felix-d:
> though I think I did find the test case for it:

@amirrustam this test case does not cover array elements hehe

Thanks @Frizi !!

Frizi:
I already have a test case for it and I'm working on a fix. Seems like this combined with other bugs deserves a full rewrite :D but don't worry, I'm not going to change public API.

Frizi:
Fixed in v0.4.0

amirrustam:
Thanks for the fix @Frizi. Will test it out today.

Issue #79, opened 2017-02-23
Error when validating a collection after removing an item


theosherman:
This is similar or perhaps a duplicate to #73. But after reading that the issue was likely resolved, I tried using the latest repo and my issue is still occurring.

There is a issue with collection validating. I will try my best to explain how to reproduce this...
I am creating a very simple example form that involves a nested collection. It's a collection of contacts, and each contact has a collection of phone numbers nested inside. It's a list of lists.

The data...
```
form: {
  "contacts": [
    {
      "name": "John",
      "phoneNumbers": [
        {
          "phoneNumber": "(111) 111-1111",
          "type": "Home"
        }
      ]
    },
    {
      "name": "Luke",
      "phoneNumbers": [
        {
          "phoneNumber": "(222) 222-2222",
          "type": "Home"
        },
        {
          "phoneNumber": "(333) 333-3333",
          "type": "Work"
        }
      ]
    }
  ]
}
```

The validation logic...
```
validations: {
    form: {
      contacts: {
        $each: {
          name: { required, minLength: minLength(4) },
          phoneNumbers: {
            $each: {
              phoneNumber: { required }
            }
          }
        }
      }
    }
  }
```

The validation data created...
```
{
  "contacts": {
    "$each": {
      "0": {
        "name": {
          "required": true,
          "minLength": true,
          "$invalid": false,
          "$dirty": true,
          "$error": false,
          "$pending": false
        },
        "phoneNumbers": {
          "$each": {
            "0": {
              "phoneNumber": {
                "required": true,
                "$invalid": false,
                "$dirty": false,
                "$error": false,
                "$pending": false
              },
              "$invalid": false,
              "$dirty": false,
              "$error": false,
              "$pending": false
            },
            "$invalid": false,
            "$dirty": false,
            "$error": false,
            "$pending": false
          },
          "$invalid": false,
          "$dirty": false,
          "$error": false,
          "$pending": false
        },
        "$invalid": false,
        "$dirty": false,
        "$error": false,
        "$pending": false
      },
      "1": {
        "name": {
          "required": true,
          "minLength": true,
          "$invalid": false,
          "$dirty": true,
          "$error": false,
          "$pending": false
        },
        "phoneNumbers": {
          "$each": {
            "0": {
              "phoneNumber": {
                "required": true,
                "$invalid": false,
                "$dirty": false,
                "$error": false,
                "$pending": false
              },
              "$invalid": false,
              "$dirty": false,
              "$error": false,
[...]

PR #96, opened 2017-03-12
Vdom rewrite

Frizi:
Full internal rewrite, exposing the same API.
- issues #95 #79 fixed and covered with regression tests
- 100% test coverage

All inner workings are now using both VDom diffing and Watchers from Vue core. The code should be in general easier to understand.


src/index.js
```diff
@@ -1,46 +1,21 @@
...
-
 const getPath = (ctx, obj, path, fallback) => {
   if (typeof path === 'function') {
     return path.call(ctx, obj, fallback)
   }
 
   path = Array.isArray(path) ? path : path.split('.')
   for (let i = 0; i < path.length; i++) {
-    if (isObject(obj)) {
+    if (typeof obj === 'object') {
```
kernelgonnapanic: What about a null check?

Frizi: :+1:

Frizi:
> you probably need sb who understands more vuelidate internals

There is nobody yet as it's brand new 😄
Still, concepts are mostly the same, it's the validation VM construction which was rewritten to use Vue's dom diffing.

src/index.js
```diff
@@ -171,185 +47,405 @@ function makePendingAsyncVm (Vue, promise) {
...
-  const validationVm = new Vue({
-    data: {
-      dirty: false,
-      dynamicKeys
+  const ValidationRule = Vue.extend({
+    props: ['rule', 'model', 'parentModel', 'rootModel'],
+    methods: {
+      runRule (parent) {
+        // Avoid using this.parentModel to not get dependent on it.
+        // Passed as an argumend for workaround
```
kernelgonnapanic: typo

Frizi:
Released in v0.4.0

alexhernandez:
@Frizi Rewrite seems to be triggering unexpected behaviors with the validators.

Frizi:
@alexhernandez please create an issue and add an example. Hard to say what unexpected behaviour mean without any clues.

PR #116, opened 2017-03-31
moved validator initialization into beforeCreate hook

ma1oy:
Need for using in provide function
```javascript
export default {
  provide() {
    return {
      validator: this.$v
    }
  }
}
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -453,14 +453,9 @@ const validationMixin = {
     const options = this.$options
     const vals = options.validations
     if (!vals) return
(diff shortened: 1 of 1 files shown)


ma1oy:
I am not good with pull requests but if you think my changes are useful commit it yourself, please :)

ma1oy:
Ah, I'm understood why it code should be in a created hook :)

Frizi:
Yep, it's using the reactive state, which is created just before `created` hook.

PR #126, opened 2017-04-14
Support boolean values in required validator

Krusen:
This adds support for required boolean values, e.g. checkboxes.

### DIFF
--- a/src/validators/common.js
+++ b/src/validators/common.js
@@ -4,7 +4,7 @@ export { withParams }
 // "required" core, used in almost every validator to allow empty values
 export const req = value => {
   if (Array.isArray(value)) return !!value.length
-  if (value === undefined || value === null) {
+  if (value === undefined || value === null || value === false) {
     return false
   }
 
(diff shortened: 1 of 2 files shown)


Frizi:
I see it's necessary to mix well with checkboxes and I clearly missed that. Thanks for the PR!

jsfeldman:
I'm not sure how common this is, but for Yes or No radio buttons I usually set the values of the elements to true and false and this commit caused a bit of an issue for me.
It might just be me, but it seems that requiring true for a checkbox is a slightly different concept than requiring a value for a field. The former has a default value, false, and is requiring a specific value, true, the latter is requiring a value exists.
I can create a custom validator for my use case, but just thought I'd bring it up to see what you all think.

petridw:
Just found this PR after creating issue #156. I don't agree with this change for the same reason as @jsfeldman, this breaks using radio buttons with the required validator. Perhaps we can find a solution that supports both radio and checkbox inputs.

Frizi:
@petridw I'd be glad to review a proposition, but I have little time to dedicate for finding it from scratch on my own. But I agree the issue is present and it has to be resolved.

petridw:
@Frizi I'll take a look at the validator when I have some time and see if a simple approach presents itself.

PR #129, opened 2017-04-23
[SSR] Implement custom virtual validation diff to safely run in SSR mode

Frizi:


Austio:
This branch fixed the SSR proxy issue for me.

PR #155, opened 2017-06-16
Increase cell size for better fit of methods name

atilacamurca:
vuelidade-bug

See `$flattenParams` method name.


Frizi:
I can take that :) Thanks 👍

PR #197, opened 2017-10-06
Add get $errorMessage

felipesdias:
Add method $errorMessage() in component, automatic generate error messages, can be use like this:

[Site Example](http://kahximiro.ddns.net/#/)

```
import { minLength, required, email } from 'vuelidate/lib/validators'

customMessages: {
  customEmail: '|field has error |dontHasParams(will be ignored)',
  minLength: 'Min |min digits',
}

validations: {
  form: {
    email: {
      required,
      customEmail: email,
      minLength: minLength(20)
    }
  }
}

customFieldName = 'Anything'

// Message 1º input
$v.form.email1.$errorMessage(customMessages, customFieldName)

// Message 2º input
$v.form.email2.$errorMessage()

// Message 3ºinput
$v.form.email3.$errorMessage({}, customFieldName

```
Already has default messages, and your default name is the name of the field.
This method return the first error message and works similar to "Laravel Validation".
Easy internationalization with customMessages.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -132,7 +132,7 @@ const validationMethods = {
         for (let j = 1; j < params.length; j++) {
           key = params[j].split(' ')[0]
           if (key === 'field') {
-            final = final.replace('|field', fieldName || validators[i]['name'] || '|field')
+            final = final.replace('|field', fieldName || this.prop || '|field')
           } else {
             final = final.replace('|' + key, validators[i]['params'][key] || '|' + key)
           }
--- a/src/index.js
+++ b/src/index.js
@@ -162,7 +162,7 @@ const validationMethods = {
           }
         }
 
-        return final.charAt(0).toUpperCase() + final.slice(1);
+        return final.charAt(0).toUpperCase() + final.slice(1)
       }
     }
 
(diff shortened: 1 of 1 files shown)


Frizi:
Sorry but this sidesteps the whole i18n story. I explicitly decided to NOT do message generation inside vuelidate, but instead provide a mechanism to make it easy to incorporate any external library focused on solving that specific problem. If you don't need translations, it's still super easy to define your own messages and just use them. You can access `$v.form.email1.$params`, so nothing prevents you from writing a function `vMessage` that accepts a validation object, like `vMessage($v.form.email1)`. Now you can create your own `Vue.prototype.vMessage = ($v) => { ... }` and done. We don't need that in the core lib. It's super easy to solve externally in "single language" scenario, and it's very important to remain flexible when you need more than that.

felipesdias:
You really are right. Thinking about this would be interesting to have the field name inside $ v, because then you do not need to pass the field name to the recognize |field

Frizi:
You already have the `type` of used validator inside generated `$params` and also the key that you control as a user. You can do `{ required }`, but also `{ mySpecialRequiredVariant: required }`, and this name is indeed visible inside `$params`. https://jsfiddle.net/5s4604ws/

Frizi:
I'm going to close that PR. If you feel something should be discussed more on that matter, feel free to open an issue.

PR #204, opened 2017-10-10
Allow setting array to undefined and null when using $each

mokkabonna:
https://jsfiddle.net/b5v4faqf/358/

It fails here: https://github.com/monterail/vuelidate/blob/master/src/index.js#L304

Since calling Object.keys on undefined or null throws.
```
 TypeError: Cannot convert undefined or null to object
      at Function.keys (<anonymous>)
      at VueComponent.keys (node_modules\vuelidate\lib\index.js:351:23)
```

Haven't been able to clone and test your code locally yet due to some computer problems, but in my code this solution behaves as expected. So just relying on circleci to run the tests for me.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -310,7 +310,7 @@ const getComponent = (Vue) => {
     computed: {
       keys () {
         var model = this.getModel()
-        if (isFunction(model) || (isObject(model) && model !== null)) {
+        if (isObject(model)) {
           return Object.keys(model)
         } else {
           return []
--- a/src/index.js
+++ b/src/index.js
@@ -301,7 +301,9 @@ const getComponent = (Vue) => {
   const EachValidation = Validation.extend({
     computed: {
       keys () {
-        return Object.keys(this.getModel())
+        var model = this.getModel()
+        if (model === undefined || model === null) return []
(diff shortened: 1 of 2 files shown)


mokkabonna:
Not sure why codecov reports the line as untouched https://codecov.io/gh/monterail/vuelidate/pull/204/diff#D1-306

It certainly get there for me in my tests. I created a dummy PR to run checks on circle CI etc. https://github.com/monterail/vuelidate/pull/204

mokkabonna:
Seems it didn't actually fail when I removed the fix. Realized I never called any computed so that the keys computed never got revaluated. Also discovered that chrome allows calling Object keys on all values but undefined and null, while phantomjs failed on some of the other values in the test. So added a more complete check for only objects. Same check lodash uses: https://github.com/lodash/lodash/blob/master/isObject.js#L26

Frizi:
Hi, thanks for that contribution! :)
Can you squash the changes? It's a whooping 8 commits that mostly overwrite each other.
Other than that, looks great!

mokkabonna:
Yeah I edited on github directly so became a lot of commits :) Thanks!

Issue #201, opened 2017-10-09
minValue & maxValue not functional with dates


mattixittam:
The minValue and maxValue validators don't seem to take into account dates. Seems a few casts would suffice as a fix:

old
`value => !req(value) || (!/\s/.test(value) && value >= min)`

new
`value => !req(value) || (!/\s/.test(+value) && +value >= +min)`

Frizi:
Hey, that's reasonable :) I would be very happy if you can provide a PR for that.

mattixittam:
Yup, will do.

Frizi:
Thanks, merged :) closing

PR #207, opened 2017-10-12
Fixed maxValue, minValue, between validators for use with Dates

mattixittam:
See issue #201

Fixed **maxValue**, **minValue** for use with Dates. Added tests, amended docs

Added tests for Date values for the **between** validator, fixed bug these turned up where all dates were invalidated due to a stringified Date containing spaces. Amended docs.


Frizi:
Hey, thanks for the contribution, and for extra *between* tests 👍  Looks good, merging :)

Issue #206, opened 2017-10-11
macAddress validator not able to validate EUI-48/MAC-48 addresses


ashleyhood:
The macAddress validator works well if the MAC address is formed as an EUI-64 address but not if it is formed as either an EUI-48 or a MAC-48 addresses.

Seeing as the IEEE has a target lifetime of 100 years for applications using MAC-48 space, this should also be supported.

All that should be required is to change line 16 of `validators/macAddress.js` to:

```javascript
return parts !== null && (parts.length === 8 || parts.length === 6) && parts.every(hexValid)
```

Frizi:
Hi, good find 👍  As you already know what to change, could you provide a PR for that? Tests are missing from your example, but that should be straightforward to add :)

Frizi:
PR merged, closing

PR #208, opened 2017-10-12
Fix validator macAddress for EUI-48 address space

ashleyhood:
As per issue #206


Frizi:
Great, thanks for that catch and great PR 👍

PR #209, opened 2017-10-12
Do not fail if parent object is non object

mokkabonna:
Similar to https://github.com/monterail/vuelidate/pull/204 vuelidate fails when parent object is set to non object. Without this fix vuelidate will fail in the test I wrote, trying to access key from undefined.

### DIFF
--- a/test/unit/specs/validation.spec.js
+++ b/test/unit/specs/validation.spec.js
@@ -9,6 +9,8 @@ const isOdd = withParams({ type: 'isOdd' }, (v) => {
   return v % 2 === 1
 })
 
+const noUndef = withParams({type: 'noUndef'}, (v) => v !== undefined)
+
 const T = () => true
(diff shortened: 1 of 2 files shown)


test/unit/specs/validation.spec.js
```diff
@@ -587,6 +587,43 @@ describe('Validation plugin', () => {
       expect(vm.$v.list.$each[0]).to.exist
       expect(vm.$v.list.$each[1]).to.not.exist
     })
+    it('should allow parent object to be non object', function () {
+      const noUndef = withParams({type: 'noUndef'}, (v) => v !== undefined)
```
Frizi: This could be moved out of test scope, near to `isOdd`.

mokkabonna:
Amended commit.

Frizi:
Thanks very much again :)

PR #221, opened 2017-11-22
Fix URL validator bug

### LINKED ISSUE #220: URL validator fails when URL is uppercase
see fiddle https://jsfiddle.net/b5v4faqf/446/

Will fail if the URL entered is in all uppercase. Simple fix would be to make it case insensitive match

JacobBennett:
Fixes #220


Frizi:
I don't think case insensitve regex is a good idea. Mixed casing in protocols is one reason. See comments on #220 for details. This fix will also need some test coverage for sure.

JacobBennett:
Updated #220 with additional details.

Frizi:
I'm convinced it's a good solution. The only missing part is some tests. If you can add at least one upper case and mixed case scenario, that would be perfect.

Frizi:
Looks perfect, thanks!

JacobBennett:
Thanks!!

PR #231, opened 2017-12-13
optimized implementation of alpha/numeric/alphanumeric validators

rkoval:
## tl;dr
optimized alpha/numeric/alphanumeric validators; see https://jsperf.com/alphanumeric-charcode-vs-regexp for benchmarks

## longer:

This PR is basically the result of a pedantic discussion I had with a coworker (sup @fortytw2). We were talking about an older repository in our organization that didn't import a validation library such as this one, and the validations done were hand-written. The gist of the rest of the discussion was this: 

_In the absence of a validation lib such as this one that abstracted away the actual validation for you_, it's not as practical to write a character code comparison validator in your application itself for a trivial validation such as alpha/numeric/alphanumeric text. You could get by with using a simple regex, much like this library has done. Although not as performant as going a bit lower level, it is easy to implement, less code to write, and overall more readable. _However_, since we _do_ have access to nice libraries such as vuelidate, the consumers of said libraries shouldn't necessarily need to care about the underlying implementation of the APIs as long as it is well tested and does what it's supposed to do. The complexity of writing a more verbose validator is effectively shielded from the rest of the world, _and_ you can get nice performance improvements as a bonus.

At any rate, I figured that since the code was easy enough to plug in and that this library is well tested, I would send a PR for it to see if the maintainers would benefit from something like this. Hopefully, you agree with the basis of the discussion that I needlessly babbled on about above such that you'd accept this change. Please let me know if you have any questions about it—it feels pretty straightforward to me.

(The benchmarks I ran are taken from https://jsperf.com/alphanumeric-charcode-vs-regexp, which has been around a while. I otherwise have nothing to do with the snippets there, so we should thank the maintainers of it for keeping it going 😄 )

### DIFF
--- a/src/validators/alpha.js
+++ b/src/validators/alpha.js
@@ -1,2 +1,17 @@
-import {regex} from './common'
-export default regex('alpha', /^[a-zA-Z]*$/)
+// taken from https://jsperf.com/alphanumeric-charcode-vs-regexp
+export default function isAlpha (str) {
+  if (!str) {
+    return true
+  }
+  var code
+
+  for (var i = 0, len = str.length; i < len; i++) {
+    code = str.charCodeAt(i)
+    if (!(code > 64 && code < 91) && // upper alpha (A-Z)
+        !(code > 96 && code < 123)) { // lower alpha (a-z)
+      return false
+    }
+  }
+  return true
+}
+
--- a/src/validators/alphaNum.js
+++ b/src/validators/alphaNum.js
@@ -1,2 +1,17 @@
-import {regex} from './common'
-export default regex('alphaNum', /^[a-zA-Z0-9]*$/)
+// taken from https://jsperf.com/alphanumeric-charcode-vs-regexp
+export default function isAlphaNumeric (str) {
+  if (!str) {
+    return true
+  }
+  var code
+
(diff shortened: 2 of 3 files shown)


Frizi:
Hi! Thanks for looking at this. I have my doubts there though.

First of all, Vuelidate is a validation engine more than a collection of validators. It is mostly optimised to run the provided validation functions as rarely as possible.

That means that if there is even a 100% performance gain on that validation function and now it is able to execute 40 million times per second, you just saved yourself at best another 1/40mln of a second every key press. That's very negligible.

I'm also very scpetical about microbenchmarks. There are so many details inside our dear javascript virtual matchines, that testing things in an environment where a single functions is rapidly evaluted alone with the exact same input is not very trustworthy. I'm sure that V8 is assuming a string type as an input, maybe even a certain internal representation of it. The compiled regex on the other hand calls to C code and deals with those cases without a possiblity of bailout (deoptimization and recompilation). 

You also dropped the `withParams` part of those validators. I actually noticed that this is not covered by unit tests.

rkoval:
This all makes sense. I'm not going to push it any further because like, you're saying, it's all pretty negligible. I just figured I'd give it a shot to see what you'd say. I appreciate you taking the time to look at the PR though!

PR #232, opened 2017-12-15
Use Puppeteer

38elements:
This replaces PhantomJS to Puppeteer.


Frizi:
This is awesome 😄 Thanks!

38elements:
[Puppeteer requires at least Node v6.4.0](https://github.com/GoogleChrome/puppeteer#installation).
I think there is no problem in here.

Frizi:
That's fine, testing suite is not critical in terms of used engine. I actually prefer using as new as possible for this kind of jobs.

Issue #136, opened 2017-04-27
Validator 'numeric' only accepts positive integers


SirLamer:
The name of the validator implies it is looking for any numeric value, including decimal and negative values. The current implementation only accepts positive integers. Either the accepted validation text should be broadened or this validator should be renamed.

Example: https://jsfiddle.net/gregpeden/hmcLdc5a/

I could provide a PR for this (fix this + provide actual integer test) if there is interest in changing it.

Frizi:
That's true. I think having a few properly named separate validations for that should do the trick. Changing that slightly breaks the API though, but we are still at a version 0 for a reason (hopefully this will change soon). A PR would be greatly appreciated :)

SirLamer:
I think it's fine to expect developers to combo "decimal + between/min/max" and "integer + between/min/max" for those who need it, we don't need lots of number format tests.

Frizi:
About the number formats, remember that we are validating model. The formatting can always be taken care of on the view layer separately, so your data variable will always hold just a plain number. But that's only one of ways to to it. Your proposal seems very reasonable. 
How i see it:
- `integer`: current numeric with support for negative values, should also accept and verify data of type `Number` to be actually integer
- `decimal`: any valid number format (anything that passes parseFloat i guess)
- `atLeast`: `min => withParams({type: 'atLeast', min}, val => val >= min`
- `atMost`: `max => withParams({type: 'atMost', max}, val => val <= max`

Nothing prevent you to use the latter one with strings, like

```javascript
{
  singleChar: maxLength(2),
  minChar: atLeast('😀'),
  maxChar: atMost('😎')
}
```
because why not ¯\\\_(ツ)\_/¯

SirLamer:
Personally I would like to copy the validation pattern set by Laravel. Using the same validation test names and input parameters has some advantages, for example implementation of the error feedback messages with localization, using the test name to perform the lookup and using any one of many "export language files to Javascript" packages with Laravel to seed the Javascript localization package from the Laravel server-side set.

https://laravel.com/docs/5.4/validation#available-validation-rules

SirLamer:
I can't get the development environment to run in Windows 10. During "npm install" I get an error "build is not recognized as an internal or external command". This happens after the npm download is done, however when I run "npm run dev" the lint testing flips out and it seems to just give up, then I can't use the test environment. So..... kinda hard to prepare a PR, then. ;-)

SirLamer:
[...]

PR #245, opened 2018-01-30
Improve numeric validator

dsbert:
As discussed in #136, numeric currently only accepts positive integer values. This pull request makes the following changes.

- Split numeric validator into numeric and integer validators.
- Update numeric validator to allow positive and negative decimal numbers.
- Limit integer validator to positive or negative whole numbers.
- Numeric validator still allows integers.
- Add unit tests for new validations.

### DIFF
--- a/src/validators/decimal.js
+++ b/src/validators/decimal.js
@@ -1,2 +1,2 @@
 import { regex } from './common'
-export default regex('numeric', /^-?[0-9.]*$/)
+export default regex('decimal', /^[-]?\d*(\.\d+)?$/)
--- a/src/validators/decimal.js
+++ b/src/validators/decimal.js
@@ -1,0 +1,2 @@
+import { regex } from './common'
+export default regex('numeric', /^-?[0-9.]*$/)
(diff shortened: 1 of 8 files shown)


Frizi:
I'm all in for your `integer` validator, but i have a few concerns about the rest. Usually "numeric" name refers to `[0-9]` character set, a string of characters composed only from digits. I'd like to keep it that way. Another argument is to keep it backward compatible if there is no reason to do otherwise. What you proposed with `/^-?[0-9.]*$/` is still valid, it just needs another name. What about `float` or `decimal`?

dsbert:
@Frizi 

Makes sense to me, I've updated the code.

src/validators/decimal.js
```diff
@@ -0,0 +1,2 @@
+import { regex } from './common'
+export default regex('numeric', /^-?[0-9.]*$/)
```
38elements: I think that it is necessary to change 'numeric' to 'decimal'.

I think that there is a bug, as below.

```js
/^-?[0-9.]*$/.test('..')
// true
```

dsbert: Thanks for the feedback. I updated the regex to disallow `..`, `123.4.5`, and `123.`

Frizi:
This is a fine pull request. Thank you @dsbert for contribution!

dsbert:
I can create a new pull request to add the validators to `index.js`. However, I think @Frizi will need to publish the updated documentation.

AlexDubok:
@dsbert It seems you haven't run `npm run build` after changing the source code

philpoore:
+1

PR #247, opened 2018-02-02
Update build scripts to enable development on Windows

dsbert:
This project does not currently build on a Windows environment as identified by @SirLamer in https://github.com/monterail/vuelidate/issues/136#issuecomment-298050082. These changes fix that.

- Add `cross-env` package to set environment variables, since `VAR=val` is invalid on Windows.
- Add `rimraf` package and use it instead of `rm -rf` which does not exist on Windows.
- Update the `exclude` path for eslint in `webpack.base.conf.js`, since Windows uses a different path delimiter.

### DIFF
--- a/build/webpack.base.conf.js
+++ b/build/webpack.base.conf.js
@@ -46,7 +46,7 @@ module.exports = {
         test: /\.js$/,
         loader: 'eslint',
         include: projectRoot,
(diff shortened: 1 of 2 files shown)


Frizi:
Thanks!

PR #261, opened 2018-03-02
Add integer and decimal validators to index.js

dsbert:
Forgot to add `integer` and `decimal` to `index.js` in this previous pull request https://github.com/monterail/vuelidate/pull/245


Frizi:
Oh, I should have noticed that. Thanks for correcting this 😄

DrCord:
When will there be a release that includes "integer" and "decimal" validators?

DrCord:
@antholord not sure why/how that effects this, but these validations are needed and based on this thread seem to be coming in the future...

Issue #175, opened 2017-08-24
Examples/documentation for using vuelidate with TypeScript


jawnsy:
Hello!

First off, thank you to the Vuelidate creators for making this amazing little validation library. Second, thank you to @mrellipse for contributing type definitions in #47.

I have been trying to figure out how to use Vuelidate with TypeScript from the PR history, by perusing the PR code, and by just generally futzing around. I see that the changes should extend the Vue interface declaration with enough information to know that the `validations` field is an available option.

However, I can't for the life of me figure out what I need to do to make this all play nicely. I tried using the `@Component` syntax from the tests (component.spec.ts) but get this error:

> ERROR in [...]\src\calculator\Calculator.vue.ts
>(96,5): error TS2345: Argument of type '{ mixins: any[]; name: string; template: string; validations: any; }' is not assignable to parameter of type 'VueClass'.
>  Object literal may only specify known properties, and 'mixins' does not exist in type 'VueClass'.

corresponding to this code (mostly borrowed from the test file, and embedded in a `.vue` single file component):

`<script lang="ts">`
```ts
import Vue = require('vue');
import Component from 'vue-class-component';
import { IVuelidate, ValidationRuleset, Vuelidate, validationMixin } from 'vuelidate';

interface IComplexProp {
    value1: string;
    value2?: number;
}

type SimpleProp = string;

interface IComponentData {
    simpleProp: SimpleProp;
    complexProp: IComplexProp;
    simpleCollection: Array<SimpleProp>;
    complexCollection: Array<IComplexProp>;
    validationGroup: string[];
};

function predicate(value: any): boolean {
    return value ? true : false;
}

function predicateFactory(param: number): (value: any) => boolean {
    return (value) => predicate(value);
}

let validations: ValidationRuleset<IComponentData> = {
    simpleProp: {
        predicate
    },
    complexProp: {
        value1: {
            predicate,
            funcGen: predicateFactory(1)
        },
        value2: {
            predicate,
            funcGen: predicateFactory(1)
        }
    },
    simpleCollection: {
        $each: {
            predicate
        }
    },
    complexCollection: {
        $each: {
            value1: {
                predicate,
                funcGen: predicateFactory(1)
            }
        }
    },
    validationGroup: ['simpleProp', 'complexProp.value2']
};

@Component({
    mixins: [validationMixin],
    name: 'RulesetTest',
    template: '<div></div>',
    validations: validations
})
[...]

PR #287, opened 2018-05-12
@types/vuelidate

janesser:
Provide types as inspired by #175.

Main types:
- Vuelidate = PluginFunction for Vue.use(Vuelidate)
- ComponentOptions.validations
- Vue.$v

Some of the predefined predicates are covered, may be not all.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,0 +1,12 @@
+// TypeScript Version: 2.4
+
+import "./vue"
+import "./lib/validators"
+
+import { Vuelidate } from "./vuelidate"
+export default Vuelidate
+
+export {
+    validationMixin,
+    Validation
+} from "./vuelidate"
--- a/types/lib/validators.d.ts
+++ b/types/lib/validators.d.ts
@@ -1,0 +1,50 @@
+// build your own rule
+declare module 'vuelidate/lib/validators' {
+    import Vue from 'vue'
+
+    export interface Params {
+        [key: string]: any,
+        type: string
+    }
+
+    export interface ValidationParams {
+        readonly name: string
+        readonly params: Params
+        readonly path: string[]
+    }
+
+    // const ValidationRule
+    export interface ValidationRule {
+        $params(): ValidationParams
+        $pending(): boolean
+    }
+
+    export type CustomRule = (value: any, parentVm?: Vue) => boolean
+
+    export interface Helpers {
+        withParams(params: Params, rule: CustomRule): ValidationRule
+        req(field: string): ValidationRule
+    }
+
+    export const helpers: Helpers
+
+    // pre-defined rules
+    export function required(): ValidationRule
+    export function requiredIf(field: string): ValidationRule
+    export function requiredUnless(field: string): ValidationRule
+    export function minLength(length: number): ValidationRule
+    export function maxLength(length: number): ValidationRule
+    export function minValue(min: number): ValidationRule
+    export function maxValue(max: number): ValidationRule
+    export function between(min: number, max: number): ValidationRule
+    export function alpha(): ValidationRule
+    export function alphaNum(): ValidationRule
+    export function numeric(): ValidationRule
+    export function email(): ValidationRule
+    export function ipAddress(): ValidationRule
+    export function macAddress(): ValidationRule
+    export function sameAs(field: string): ValidationRule
+    export function url(): ValidationRule
+    export function or(...validators: ValidationRule[]): ValidationRule
+    export function and(...validators: ValidationRule[]): ValidationRule
+}
--- a/types/vue.d.ts
+++ b/types/vue.d.ts
@@ -1,0 +1,49 @@
+// vue augmentation
+import Vue, { ComponentOptions } from 'vue'
+
+import { ValidationRule } from 'vuelidate/lib/validators'
+import { Validation } from "./vuelidate"
+
+interface DynamicValidation<V extends Vue> { }
+
+type ValidationProperties<V extends Vue> = {
+    [P in keyof V]?: Validation
+}
+
+interface ValidationGroups<V extends Vue> {
+    [groupName: string]: ValidationProperties<V>
+}
+
+declare module 'vue/types/vue' {
+    interface Vue {
+        $v: ValidationProperties<this> & ValidationGroups<this> & Validation
+
+        delayTouch(v: Validation): void
+
+        validations(): DynamicValidation<this>
+    }
+}
+
+interface ValidGroupDecl {
+    [group: string]: string[]
+}
+
+
+interface ValidPropertyDecl {
+    [prop: string]: RuleDecl
+}
+
+type ValidationDecl = ValidationRule | ((...args: any[]) => ValidationRule)
+type GroupDecl = string[]
+type AsyncDecl = (...args: any[]) => boolean | Promise<boolean>
+type NestedDecl = RuleDecl
+interface RuleDecl {
+    [rule: string]: ValidationDecl | GroupDecl | AsyncDecl | NestedDecl
+}
+
+
+declare module 'vue/types/options' {
+    interface ComponentOptions<V extends Vue> {
+        validations?: RuleDecl
+    }
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -380,6 +380,28 @@
         "chalk": "^2.0.0",
         "esutils": "^2.0.2",
         "js-tokens": "^3.0.0"
+      },
+      "dependencies": {
+        "ansi-styles": {
+          "version": "3.2.1",
+          "resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-3.2.1.tgz",
+          "integrity": "sha512-VT0ZI6kZRdTh8YyJw3SMbYm/u+NqfsAxEpWO0Pf9sq8/e94WxxOpPKx9FR1FlyCtOVDNOQ+8ntlqFxiRc+r5qA==",
+          "dev": true,
+          "requires": {
+            "color-convert": "^1.9.0"
+          }
+        },
+        "chalk": {
+          "version": "2.4.1",
+          "resolved": "https://registry.npmjs.org/chalk/-/chalk-2.4.1.tgz",
+          "integrity": "sha512-ObN6h1v2fTJSmUXoS3nMQ92LbDK9be4TV+6G+omQlGJFdcUX5heKi1LZ1YnRMIgwTLEj3E24bT6tYni50rlCfQ==",
+          "dev": true,
+          "requires": {
+            "ansi-styles": "^3.2.1",
+            "escape-string-regexp": "^1.0.5",
+            "supports-color": "^5.3.0"
+          }
+        }
       }
     },
     "@babel/plugin-proposal-async-generator-functions": {
@@ -877,10 +899,16 @@
         "@types/babel-types": "*"
       }
     },
+    "@types/parsimmon": {
+      "version": "1.10.0",
+      "resolved": "https://registry.npmjs.org/@types/parsimmon/-/parsimmon-1.10.0.tgz",
+      "integrity": "sha512-bsTIJFVQv7jnvNiC42ld2pQW2KRI+pAG243L+iATvqzy3X6+NH1obz2itRKDZZ8VVhN3wjwYax/VBGCcXzgTqQ==",
+      "dev": true
+    },
     "@vue/component-compiler-utils": {
-      "version": "1.2.1",
-      "resolved": "https://registry.npmjs.org/@vue/component-compiler-utils/-/component-compiler-utils-1.2.1.tgz",
-      "integrity": "sha512-l3GdahBgXlp/SoY5KU7mqMqg/BNiJAndpw/6nMfGzFooCGUkq49CT3pCMiSYy0g/2a6iBD37cHBYlsg4nulIwQ==",
+      "version": "1.3.0",
+      "resolved": "https://registry.npmjs.org/@vue/component-compiler-utils/-/component-compiler-utils-1.3.0.tgz",
+      "integrity": "sha512-CerIPMzE6y4J4v68nUh2NFPjOO8zr2r8UDh/yi63E4bf7YB5Aqhm7Fv90zzTF+sOQgsqTb9dgryvpeDY1/7m9g==",
       "dev": true,
       "requires": {
         "consolidate": "^0.15.1",
@@ -1121,9 +1149,9 @@
       }
     },
     "abbrev": {
-      "version": "1.0.9",
-      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.0.9.tgz",
-      "integrity": "sha1-kbR5JYinc4wl813W9jdSovh3YTU=",
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
+      "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q==",
       "dev": true
     },
     "accepts": {
@@ -1321,13 +1349,16 @@
       "dev": true
     },
     "ansi-styles": {
-      "version": "3.2.1",
-      "resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-3.2.1.tgz",
-      "integrity": "sha512-VT0ZI6kZRdTh8YyJw3SMbYm/u+NqfsAxEpWO0Pf9sq8/e94WxxOpPKx9FR1FlyCtOVDNOQ+8ntlqFxiRc+r5qA==",
-      "dev": true,
-      "requires": {
-        "color-convert": "^1.9.0"
-      }
+      "version": "2.2.1",
+      "resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-2.2.1.tgz",
+      "integrity": "sha1-tDLdM1i2NM914eRmQ2gkBTPB3b4=",
+      "dev": true
+    },
+    "any-promise": {
+      "version": "1.3.0",
+      "resolved": "https://registry.npmjs.org/any-promise/-/any-promise-1.3.0.tgz",
+      "integrity": "sha1-q8av7tzqUugJzcA3au0845Y10X8=",
+      "dev": true
     },
     "anymatch": {
       "version": "2.0.0",
@@ -1595,33 +1626,6 @@
         "chalk": "^1.1.3",
         "esutils": "^2.0.2",
         "js-tokens": "^3.0.2"
-      },
-      "dependencies": {
-        "ansi-styles": {
-          "version": "2.2.1",
-          "resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-2.2.1.tgz",
-          "integrity": "sha1-tDLdM1i2NM914eRmQ2gkBTPB3b4=",
-          "dev": true
-        },
-        "chalk": {
-          "version": "1.1.3",
-          "resolved": "https://registry.npmjs.org/chalk/-/chalk-1.1.3.tgz",
-          "integrity": "sha1-qBFcVeSnAv5NFQq9OHKCKn4J/Jg=",
-          "dev": true,
-          "requires": {
-            "ansi-styles": "^2.2.1",
-            "escape-string-regexp": "^1.0.2",
-            "has-ansi": "^2.0.0",
-            "strip-ansi": "^3.0.0",
-            "supports-color": "^2.0.0"
-          }
-        },
-        "supports-color": {
-          "version": "2.0.0",
-          "resolved": "https://registry.npmjs.org/supports-color/-/supports-color-2.0.0.tgz",
-          "integrity": "sha1-U10EXOa2Nj+kARcIRimZXp3zJMc=",
-          "dev": true
-        }
       }
     },
     "babel-generator": {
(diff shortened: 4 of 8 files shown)


archSeer:
I get a few errors with this + class based components:

```
57:5 Cannot invoke an expression whose type lacks a call signature. Type 'Validation' has no compatible call signatures.
    55 |   public register() {
    56 |     // force the form to validate
  > 57 |     this.$v.$touch();
```

Also, an issue I was having even with the typings provided in the linked issue (I'm using typescript 2.8):
```
@Component({
  mixins: [validationMixin],
  components: {
    ...
  },
  validations: {
    model: {
      name: { required },
    },
  }
})

```

```
32:3 Argument of type '{ mixins: any[]; components: { ...' is not assignable to parameter of type 'VueClass<Vue>'.
  Object literal may only specify known properties, but 'mixins' does not exist in type 'VueClass<Vue>'. Did you mean to write 'mixin'?
```

It will always error on the first key inside the object, but the error is actually for `validations`. If I remove all validations from the component, it starts passing again.

I was using this, but it doesn't fix it at all:

```ts
declare module 'vuelidate' {
  // we need to augment the actual Vue types
  import vue from 'vue'

  // must bolt on the validations property to the input of ComponentOptions
  // and also onto the "output" of @Component in the form of VueClass<Vue>
  module 'vue/types/options' {
    interface ComponentOptions<V extends vue> {
      validations?: { [x: string]: any }
    }
    interface VueClass<V extends vue> {
      validations?: { [x: string]: any }
    }
  }
```
https://github.com/vuejs/vue-class-component/issues/100#issuecomment-357154506

janesser:
i can post my example-usage project next days, so we can compare setup more precisely.

from first glance there is missing:
- validationGroups needs recursive type.
- this.$v.$touch() i didn't expect this one without an attribute,
let me know if you are serious about it.

archSeer:
So, I installed the playground project and that `@Component()` error happens in there as well:

```
[ts]
Argument of type '{ template: any; components: { 'b-container': any; 'b-col': any; 'b-row': any; }; validations: { ...' is not assignable to parameter of type 'VueClass<Vue>'.
  Object literal may only specify known properties, and 'template' does not exist in type 'VueClass<Vue>'.
```

screen shot 2018-05-16 at 12 46 03

Trying to see if it's really a typescript version issue by downgrading to 2.4

janesser:
Could be occurrence of this one: https://github.com/vuejs/vue-class-component/issues/183

Actually i was stuck with a similar error until i re-bootstrupped with 
https://github.com/ducksoupdev/vue-webpack-typescript
(but this is ts2.7)

what's your current version?

janesser:
Hey there, sorry for the delay.

I completed the missing parts of validationGroup etc..
Now vue-playground (link above) contains all examples from https://monterail.github.io/vuelidate/.

You'd better clean check-out. npm cheats on me sometimes.

Let me know if further issues arise.

Otherwise let's merge it.

types/index.d.ts
```diff
@@ -0,0 +1,101 @@
...
+    readonly $invalid: boolean
+    readonly $dirty: boolean
+    readonly $error: boolean
+    readonly $pending: boolean
+    readonly $params: { [attr: string]: any }
+
+    // const validationMethods
+    $touch(): never
+    $reset(): never
+    $flattenparams(): Params[]
```
cypherix93: This seems to be a typo. Should be `$flattenParams()`.

See here: https://github.com/monterail/vuelidate/blob/a6e74f4d2b8c16a3396b1fe141729412fcd5cf0d/src/index.js#L135

Also looks like `Params` is not the correct type for this. I am getting back an object that contains the `params: Params`, along with `name` and `path`.

Screenshot: 
screenshot_256

janesser: adjusted types accordingly.

janesser:
rebased on 3a8ae66cb59433cd00b793668fde3ff3f62ceff0

types/index.d.ts
janesser: Thank you @cypherix93.
I fixed the mistake in the PR.

About "Params" i share your doubt. It is coupled to withParams(..) and i have seen examples where there is more than the "type" attribute. Besides it has changed lately.

Seems like the whole 
"import { helpers } from 'vuelidate/lib/validators'"
is new and actually uncovered.

jubairsaidi:
This is coming soon I hope?

laland:
Guys? Will you merge it?

maciz84:
This would be really nice

backbone87:
I wanted to try this branch.
-> `npm i monterail/vuelidate#pull/287/head`

But I get an error from webpack (while VS Code says everything is fine):
`This dependency was not found: vuelidate in ./src/vuelidate.ts`

Any idea?

janesser:
@backbone87 i can reproduce, using "npm i monterail/vuelidate#pull/287/head".

The way i have it working is:
in clone of this PR, execute: "npm pack"
in my vue-playground, "npm i ../vuelidate/vulidate-0.7.3.tgz"

I guess that "npm build" isn't executed, when cloning directly into node_modules. Let me try.

backbone87:
i think you are correct, i tried to `npm run build` after installing the PR via npm, but this doesnt work since most config files are .npmignore'd. so i clone the PR to local, build it and npm installed it from local

however your typings dont work for me (TS 2.9). I ended up copying your definitions into my project and made some changes (and installed latest vuelidate from npm)

here are typings i use:
https://gist.github.com/backbone87/dc1f3a981207d6d435f5209208637786

janesser:
Rebased on vuelidate 0.7.4.
Upgraded to latest typescript 2.9.2.

Included input from @backbone87. Thanks man.
Updated vue-playground.

@Frizi It would be about time to merge. What does your plan look like?

backbone87:
ok i got it working from local.

after checkout i just did `npm i ../vuelidate` which linked the local repo into my projects node_modules. this somehow screwed up the type augmentation because the local repo had its own node_modules.

the solution was `npm pack` and installing the tarball instead.

these typings work for me now so far.

janesser:
@backbone87 thx for pushing the point. I reworked the structure. Have a look, if you don't mind. It's pretty similar like vue-router.

@Frizi Still awaiting your feedback. Make a move!

types/index.d.ts
```diff
@@ -0,0 +1,9 @@
+// TypeScript Version: 2.4
+
+import "./vue"
+
+import { Vuelidate } from 'vuelidate/types/vuelidate'
```
backbone87: if you leave out the `declare module 'vuelidate/types/vuelidate'` from the `vuelidate.d.ts` then you can just do `import ... from 'vuelidate'`

wagich:
Shouldn't the `CustomRule` type be included when defining `ValidationDecl` in vue.d.ts?

Additionally, I was able to get (all) instance properties to show up on the `$v` property, enabling type checking and IntelliSense.
```typescript
// vue.d.ts
type ValidationProperties<V extends Vue> = {
    [P in keyof V]?: Validation;
}

declare module 'vue/types/vue' {
    interface Vue {
        $v: ValidationProperties<this> & Validation
        delayTouch(v: Validation): void
        validations(): ValidationDecls
    }
}
```
I was also able to flow the component type to a custom validator by adding a generic type parameter to the `ValidationDecl*` interfaces. Gist with all changes: https://gist.github.com/wagich/aa0b847c7ab45c61bd11d6cb0d16e4c1

At the moment this is only tested using class-style components and pretty simple validations but it could potentially be useful.

types/index.d.ts
```diff
@@ -0,0 +1,9 @@
+// TypeScript Version: 2.4
```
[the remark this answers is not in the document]
janesser: dtslint - there are issues with typescript 2.3.

I would assume these have no relevance any more.

types/vue.d.ts
```diff
@@ -0,0 +1,41 @@
+// vue augmentation
+import Vue, { ComponentOptions } from 'vue'
+
+import { ValidationRule } from 'vuelidate/lib/validators'
```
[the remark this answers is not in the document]
janesser: the typing with vue is hardly related.

types/lib/validators.d.ts
```diff
@@ -0,0 +1,49 @@
+// build your own rule
+declare module 'vuelidate/lib/validators' {
+    import Vue from 'vue'
+
+    export interface Params {
```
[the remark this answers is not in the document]
janesser: Thanks mate, Params interface updated.

OlivierGonthier:
Hi! Any updates on this ?

danjohnso:
Since @Frizi doesn't seem to be in a hurry...any advice in integrating this with my project directly?  I put the types int my project and then included them in my ts.config:

`"typeRoots": [
			"src/ts/types",
			"node_modules/@types"
		]`

Unfortunately, at compile time when I import vuelidate it can't seem to figure out the types:

Import:
`import Vuelidate from 'vuelidate';`

Error
`TS7016: Could not find a declaration file for module 'vuelidate'. '.\node_modules\vuelidate\lib\index.js' implicitly has an 'any' type`

janesser:
@wujekbogdan DefinetelyTyped is kinda outdated. That gave a very hard-time finding out which typings belong to which version. It would be good fit for a maintenance project, but vuelidate release 0.7.5 while this PR was open, so that typings adjustment was required.

@Frizi I understand your hesitation, still any chance to bribe you with beer or something?

@danjohnso you can have a look at vue-playground (linked somewhere above).
Is there some gist about your attempt?

TomCaserta:
Any progress on accepting this PR?

It'd be much easier to have this available with the package itself.

@janesser what do you mean by definitely typed being outdated? It's simply a repository to place these typings until the library authors integrate them into their project themselves.

Rouche:
@janesser Nice work. Would you be able to deploy your definitions in npm? So we can use `npm install 
--save-dev blabla`

wujekbogdan:
@TomCaserta

> what do you mean by definitely typed being outdated? It's simply a repository to place these typings until the library authors integrate them into their project themselves.

I think that @janesser meant that the prefered way of storing types declarations is to keep them under projects' repositories instead of the DefinetelyTyped repo because it's easy to get versions mismatch, especially when using an older version of a package. If this is what he meant, then I can't disagree... but on the other hand, it's better to have types declaration in any repo than nowhere ;)

Rouche:
I am trying to keep them in my project with
```

**Comment by Rouche** [comment]
Theres missing types on Helpers. Especially for custom validators like sameAs is using.
Here is the lest, im unsure if i can do a PR on a PR tough.

Should look like this:
```
    export interface Helpers {
        withParams(params: Params, rule: CustomRule): ValidationRule
        req(field: string): ValidationRule
        ref(reference: any, vm: Vue, parentVm: Vue): any
        len(value: any): number
        regex(type: any, expr: any): boolean //Not sure about this one
    }
```

**Comment by janesser** [comment]
@Rouche i will add the missing helpers.

@wujekbogdan @TomCaserta i can investigate in options that work without this PR being merged, though it is question of time until it runs out of sync. This will rely on support over PRs...

I plan to have a session this weekend.

**Comment by TomCaserta** [comment]
Ah yeah, good point, in any case likely I'll also end up helping keeping it up to date as and when it needs to stay in sync, at least until it's integrated into the package itself. 

Good work though and hope to see this merged sooner rather than later :) :+1:

**Comment by Rouche** [comment]
I will help too. Whenever possible. I do not think it will get out of sync. The API cannot change drastically or it will break users. Maybe missing features but i fail to see what could be added :)

**Comment by janesser** [comment]
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/29931

**Comment by rmaclean-ee** [comment]
I found this worked for regex

```js
regex(type: any, expr: RegExp): ValidationRule; 
```

chroda:
Do we have TS support already?

I wish to see a clear example of how to use, maybe a section in the docs will be amazing.

janesser:
@chroda https://github.com/monterail/vuelidate/pull/287#issuecomment-408606711 this is best available now. Actually vuelidate owners aren't reacting to this PR, so cross fingers for the DefinetelyTyped approach.

janesser:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/29931

@types/vuelidate got merged and published over night.

Thanks to all supporters. Feedback welcome.

"npm i @types/vuelidate --save-dev" to give it a try.

Rouche:
Works like a charm!!! Great job!

types/vue.d.ts
```diff
@@ -0,0 +1,49 @@
...
+type AsyncDecl = (...args: any[]) => boolean | Promise<boolean>
+type NestedDecl = RuleDecl
+interface RuleDecl {
+    [rule: string]: ValidationDecl | GroupDecl | AsyncDecl | NestedDecl
+}
+
+
+declare module 'vue/types/options' {
+    interface ComponentOptions<V extends Vue> {
+        validations?: RuleDecl
```
xSorc: With this typing i can't use [Dynamic validation schema](https://monterail.github.io/vuelidate/#sub-dynamic-validation-schema). Maybe its better to write `RuleDecl | (() => RuleDecl)`?

stas-kh:
Hi @janesser 

Thank you for the provided typings. I have some feedback on it.
If I use an object for the form like:
```
public formData = {
  userId: null,
};
```
and the following validation:
```
const validations = {
  formData: {
    userId: {
      required,
      isValid(value): boolean {
        // do something
      },
    },
  },
};
```
then I cannot `touch` this field in TypeScript because of error:
```
ERROR in %file-path%/SomeFile.vue
Property 'userId' does not exist on type 'Validation'.
    124 |     } else {
    125 |       this.formData.userId = option.id;
  > 126 |       this.$v.formData.userId.$touch();
        |                          ^
    127 |     }
    128 |   }
```

Rouche:
@stas-kh i dont think there is a way to Generic $v.formData.

You will need an interface and cast your formData

```
(this.$v.formData as MyForm).userId.$touch()
```

stas-kh:
@Rouche thank you for the reply, that's how we decided to proceed right now but it would be great to have this fixed somehow.

Because depending on [Vuelidate docs](https://monterail.github.io/vuelidate/#sub-validation-groups) that's a valid case to use validation groups

janesser:
> 
> 
> @Rouche thank you for the reply, that's how we decided to proceed right now but it would be great to have this fixed somehow.
> 
> Because depending on [Vuelidate docs](https://monterail.github.io/vuelidate/#sub-validation-groups) that's a valid case to use validation groups

@stas-kh $v.anyThing is badly typed since the typescript generics won't be able to help here.
i'll have a closer look on week-end.

types/vue.d.ts
janesser: hi @xSorc ,

i have this working example: https://github.com/janesser/vue-playground/blob/4ff484eb9f20adc832b520539124afad1c751cba/src/components/valid/valid.ts#L118

what's different about yours?

xSorc: Hi @janesser ,

Is your example realy works? You have duplicate **validations** variable. One in the component decorator and the same one in the list of functions below. And witch one is correct? I'd reproduced this case localy - function **validations** doesn't call.

I think the function **validations** doesn't calls, because you need to declare **validations** variable directly in the root vue object. But vue-property-decorator moves all you functions into the methods section. The right place for function **validations** - in the decorator.

p.s. sorry for bad grammar :)

janesser: ill have a closer look, thanks for the hints.

stas-kh:
Hi @janesser. Do you have any updates?

janesser:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/30701

@xSorc @regnisolbap regarding "dynamic validation schema" proposed improvement is included.

@stas-kh @Rouche i pulled the best out of typescript type system.

Reviews and Comments welcome.

Rouche:
Should do the job. But like i said its impossible to "generic" the underlying implementation of the form class.
Even if you ```extends Vue<MyForm>``` it wont be enough, because  there are cases where a Vue have 2 forms and you dont want to do 2 components.

Dont loose time trying. Its already good as it is.

Meanwhile, there is a bigger problems.... (vuelidate with vuetifyjs)

janesser:
> Should do the job. But like i said its impossible to "generic" the underlying implementation of the form class.
> Even if you `extends Vue<MyForm>` it wont be enough, because there are cases where a Vue have 2 forms and you dont want to do 2 components.
> 
> Dont loose time trying. Its already good as it is.
> 
> Meanwhile, there is a bigger problem.... (vuelidate with vuetifyjs)

Sounds interesting, please link PRs here.

Rouche:
@janesser Yes but also:

- ```@input="$v.name.$touch()"``` is not working. There is probably a prevent somewhere in Vuetifyjs code. Very hard to debug so i fell back to ```:rules="[]"```
- Validations does not reset after manual $touch() call on form

Anyway i abandonned trying. Ill do a JSFiddle when i have time.

janesser:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/30701 got merged.

Please give feedback about the issues handled.

regnisolbap:
Hey @janesser, thanks a lot for the updates ! However I encountered some issue with the new **ValidationProperties** type, I'm using a webpack-dev-server with ts-loader and typechecking takes a lot more time than with the old one (I tweaked it directly from node_modules and relaunched the server). Do you have any idea what this could be about ?

janesser:
@regnisolbap can you provide some gist of your use case?

regnisolbap:
@janesser, sorry for the mess, I just updated my projects typescript version from 3.1.1 to 3.2.1 and the typechecking works like before, thanks !

wujekbogdan:
@janesser

I've installed `@types/vuelidate ^0.7.2`. Here's a very basic usage example:

``` html
<script lang="ts">
import { Component, Vue } from 'vue-property-decorator';
import { validationMixin } from 'vuelidate';
import { required } from 'vuelidate/lib/validators';

@Component({
  mixins: [validationMixin],
  validations: {
    something: { required },
  },
})
export default class MyComponent extends Vue {
  something = 'test';
}
</script>
```

I'm getting the following error:
```
Argument of type '{ mixins: any[]; validations: { something: { required: () => ValidationRule; }; }; }' is not assignable to parameter of type 'VueClass<Vue>'.
  Object literal may only specify known properties, but 'mixins' does not exist in type 'VueClass<Vue>'. Did you mean to write 'mixin'?
```

janesser:
@wujekbogdan indeed the mixin isn't required to be made explicit.

Feel free to check
https://github.com/janesser/vue-playground
which demonstrates a basic way of enabling vuelidate with typescript-support.

The error message can point to some two vue-versions installed in parallel.
Actually the types mentioned do seem compatible.
You should probably revisit your "npm ls".

wujekbogdan:
@janesser 
I have deleted `node_modules` dir and `yarn.lock` file and ran `yarn install`.
It works fine now.

jubairsaidi:
@wujekbogdan, @janesser does have a good point, I'll try to make the change in the next few days.

wujekbogdan:
@janesser
I have one more problem, it's not a bug - it's rather my poor TS knowledge.

``` typescript
interface IForm {
  email: string,
  name: string,
}

@Component({
  validations: {
    form: {
      email: {
        required,
        email,
      },
      name: {
        required,
      },
    },
  },
})
export default class Form extends Vue {
  form: IForm = {
    email: '',
    name: '',
  };

  getInputClass(name) {
    if (!this.$v.form[name].$invalid) {
      return '';
    }

    return 'form__input--invalid';
  }
}
```

It's obvious to me why it happens, but I have no idea how to fix it. The following code causes TS compilation error: `Unresolved variable form`. 

``` js
if (this.$v.form[name].$invalid)
```

It happens because TS compiler doesn't know that `this.$v.form` property exists. It also doesn't know that it implements the `Validation` interface. 

I know that I need to write additional typings, but I don't know how to do it. I would appreciate any help.

guilhermelimak:
I think there's still an error with the type definitions, after installing it on my project the intellisense inside .vue files gets really slow to the point of being unusable, showing the errors on weird places, etc.

I was able to trace it to the @types/vuelidate package, more specifically on this line:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/945fd1a38258b551241eba913701b313c28b7884/types/vuelidate/vue.d.ts#L9

if I remove the `ValidationProperties<V[P]>` part it works fine (but then the types are wrong of course).

I think it's being caused because it's using the whole `V` type recursively, which might get heavy (I guess?, we do have a large amount of files on this project), I think we should only pass the validation properties there, not the whole instance, but I can't seem to access the validation properties provided by the user on that part, does anyone have any ideas on how this can be acomplished?

janesser:
> 
> 
> I think there's still an error with the type definitions, after installing it on my project the intellisense inside .vue files gets really slow to the point of being unusable, showing the errors on weird places, etc.
> 
> I was able to trace it to the @types/vuelidate package, more specifically on this line:
> https://github.com/DefinitelyTyped/DefinitelyTyped/blob/945fd1a38258b551241eba913701b313c28b7884/types/vuelidate/vue.d.ts#L9
> 
> if I remove the `ValidationProperties<V[P]>` part it works fine (but then the types are wrong of course).
> 
> I think it's being caused because it's using the whole `V` type recursively, which might get heavy (I guess?, we do have a large amount of files on this project), I think we should only pass the validation properties there, not the whole instance, but I can't seem to access the validation properties provided by the user on that part, does anyone have any ideas on how this can be acomplished?

i am working with a very old single core amd k125, and i've been okay.
can you provide some example-project, where you faced some slowness?

regards,
jan

TomCaserta:
Think I've found a problem with the types when using `strictNullTypes`. 

If you attempt to use `this.$v.someProperty!.$invalid` for example it will complain: 

```
[tsl] ERROR in /x/x.ts(73,38)
TS2339: Property '$invalid' does not exist on type 'NonNullable<Validation & ValidationProperties<this["someProperty"]>>'.
```

Maybe I'm wrong but I think theres some weird issue with the recursion of the `$v` type. I've modified the type to be 

```Typescript
        $v: ValidationProperties<Pick<this, Exclude<keyof this, '$v'>>> & ValidationGroups & Validation
```

And this seems to behave correctly now but I'm unsure if theres something else at play here that you may be able to identify. Will create a PR on definitely typed so if you could review and check that'd be great.

RobCherry:
I can second @guilhermelimak conclusion that there is an issue with type definitions.

I am seeing an out of memory error similar to https://github.com/Microsoft/TypeScript/issues/21048 when using [@types/vuelidate](https://www.npmjs.com/package/@types/vuelidate) version 0.7.2 or higher.  Using 0.7.1 appears to work fine.

RobCherry:
I take it back, that is an issue when running with typescript version 3.1.6, however after upgrading to 3.2.4 the out of memory issue appears to be resolved.

janesser:
@Frizi any news here?

jubairsaidi:
I'll try and make time for this this week! Sorry been busy!

ktiedt:
I never realized these typings with all these "great" feedback comments from maintainers got ignored all this time... I suspect this explains my difficulties getting vuelidate to work with typescript :( 

Rather sad to see such amazing work just sitting here potentially aging out of being relevant as projects continue to grow.

janesser:
i am closing this as the @types/vuelidate progresses and this PR becomes outdated.

Cheers.

PR #292, opened 2018-05-21
Implement and document $anyDirty and $anyError flag

### LINKED ISSUE #84: version of $dirty that is true if any children have been changed
I would also like to have a value that is similar to $dirty with the difference being that it is true if **any** (rather than all) of the children are $dirty. The purpose to make it easy to track if there have been any changes in the overall form or a specific section of the form - e.g. certain actions will be available if even only one value has changed.

I am just not sure how best to do this. I understand that $dirty is a computed value, so preferably I would like to add a complementary computed value to $dirty (e.g. $anydirty) to be access in a similar manner to the $dirty value. If anyone in the know has some guidance on the best way to achieve this it would be greatly appreciated.

Frizi:
closes #84

### DIFF
--- a/docs/main.js
+++ b/docs/main.js
@@ -54,7 +54,7 @@ new Vue({
       const navTop = $nav.getBoundingClientRect().top
       this.isNavSticky = navTop < 0
 
-      const found = this.allHeaders.findIndex(e => {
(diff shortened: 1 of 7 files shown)


driekus77:
Applause for this anyDirty and anyError feature! Thanks!
🥇

Issue #260, opened 2018-03-01
Document and code are not synchronized！


seasolzombie:
For example, writing according to the document, Custom Validators Section is not correct.
Following is the right code for importing Custom Validators

import {withParams} from 'vuelidate/lib/params'
export const isHello = withParams({type: 'isHello'}, value => value === 'hello')

phcoliveira:
This didn't work for me. I am kind of clueless about how to create custom validators.

Issue #253, opened 2018-02-20
Check if only letters and/ or spaces


sjardi:
Hello! I wanted to do a check where a string only contains letters or spaces ( first name and lastname ), but it gives a false if i enter a space between first and lastname. Is there a recommended way to fix this?

lawsonry:
You can always use a regex pattern, like this:

    <input type="text" pattern="[a-zA-Z0-9]+" data-error="Must be alphanumberic" data-success="Great!">

robodude:
I would also like to see a pre-made validation function provided that would check letters or numbers or spaces. I would prefer to keep all my validation 'stuff' happening in the same place.

sjard1:
Well you can make a custom validation, but the way vuelidate fix this is that you almost completly remove the vuelidate syntax and completly write it in normal javascript..

robodude:
You're absolutely correct, @sjard1 . I was only suggesting that this type of validation could be offered out of the box as one of the built in validators as seen here: 
https://monterail.github.io/vuelidate/#sub-builtin-validators

Maybe this is a good opportunity for me or someone to build one and submit a PR ;)

lawsonry:
@robodude that's a good idea! I just recently ran into this problem and ended up rolling my own quick and dirty front-end solution. I think it would be smart to offer more solutions as pre-built validators out-of-the-box.

Frizi:
Improved docs for custom validators incoming. This should resolve this issue.

JordanaCaiCarv:
Hi! This solution works. Thanks. But... If I have something like this "Aaaaaaaaaa bbbbb ccc" , how can  I trate? Thinking in security too.

sjardi:
> Hi! This solution works. Thanks. But... If I have something like this "Aaaaaaaaaa bbbbb ccc" , how can I trate? Thinking in security too.

What is trate? And what do you mean in security to? If you want custom validation you must use your own regex as jesselawson described above. The so after pattern= "Your own regex here", if you don't know what regex is or how it works you should google it ;)

good luck!

Issue #240, opened 2018-01-22
Is it possible to use `withParams` in an asynchronous validator?


samtsai:
I'm trying to create a custom validator that is asynchronous (ie loads in an external library on demand) and its important for me to append useful `$params` but I can't seem to get both async validation and `withParams` working:
https://jsfiddle.net/b5v4faqf/527/

Validation works but `params` are always `null`.

Frizi:
`withParams` is synchronous by nature. There is no way to append params inside async call. What are you trying to accomplish? Maybe there is other more natural way to do it? If not, it might be a valid case to extend the library.

samtsai:
Thanks for the response @Frizi . Albeit, I might be trying an edge case but what I'm trying to do is:

- load the Dropbox library [zxcvbn](https://github.com/dropbox/zxcvbn) asynchronously (on-demand)
- validate password strength
- attach `zxcvbn` strength messages to params via `withParams`

```
import { required } from 'vuelidate/lib/validators';
import { withParams } from 'vuelidate/lib/validators/common';

export default (requirement, value) => {
  return import(/* webpackChunkName: "zxcvbn" */ 'zxcvbn').then(zxcvbn => {
    let strength = zxcvbn(value);
    let score = strength.score;
    let isStrong = score >= requirement;
    let message;

    if (!isStrong) {
      message = gatherMessage(strength.feedback);
    }

    return withParams({
      type: 'passwordStrength',
      feedback: message,
      requirement,
      isStrong,
    }, () => {
      return required(value) && isStrong;
    });
  });
};
```

My current implementation:

```
import { required } from 'vuelidate/lib/validators';
import { withParams } from 'vuelidate/lib/validators/common';
import zxcvbn from 'zxcvbn';

export default (requirement, value) => {
  let passwordStrength = zxcvbn(value);
  let score = passwordStrength.score;
  let isStrong = score >= requirement;
  let message;

  if (!isStrong) {
    // `gatherMessage` defined outside of `export`
    message = gatherMessage(passwordStrength.feedback);
  }

  return withParams({
    type: 'passwordStrength',
    feedback: message,
    requirement,
    isStrong
  }, () => {
    return required(value) && isStrong : true;
  });
};
```

This was inspired by my reading this post: http://jmduke.com/posts/how-i-cut-my-webpack-bundle-size-in-half/

samtsai:
[...]

Issue #168, opened 2017-08-03
Custom Validator: req, withParams use in project file


franz-josef-kaiser:
We just had a situation where one of our designers needed to add a custom validator. The [`src/common.js`](https://github.com/monterail/vuelidate/blob/1f21cf649e56bed942c891c464276196e14fd240/src/validators/index.js) does not export what is shown in the [_"Custom Validators"_ Section](https://monterail.github.io/vuelidate/#sub-custom-validators) in the documentation:

```js
import {withParams, req} from './common'
export default withParams({type: 'required'}, req)
```

The `withParams` and `req` object get imported from the module with a _local_ path. How would one import it inside a project when adding validators to the project folder? Thanks!

Frizi:
Sure, that's a fail on documentation part. I never intended to make those `common` building blocks public, as those are not validators really. I refactored them out once and forgot that docs are autogenerated from the source. As you can see in the code you pasted in, `required` is an public equivalent of `req`. Also `withParams` is exported from `vuelidate/lib/validators`.

First of all, if you are not using `$params` in your application, you can skip this altogether and simply write a JS function.
```javascript
export const isHello = value => value === 'hello'
```

If you want to use meaningful `$params` with that, you have to wrap it in `withParams`.

```javascript
import {withParams} from 'vuelidate/lib/validators'
export const isHello = withParams({type: 'isHello'}, value => value === 'hello')
```

franz-josef-kaiser:
Shall we leave this open as commit and PR target until the doku is fixed?

shanemgrey:
I've looked at the example in #31 but that doesn't seem to work for me. I was able to follow the "with_params" example in #137 and get it to work, but don't really understand the need for withParams and can't find any instruction on it in the docs. 

For this one I just have a simple case where I want to check is a string is a valid date.  (This works)

```import { withParams } from 'vuelidate/lib'
import isValid from 'date-fns/is_valid'

export default withParams({ type: 'isValidDate' }, dateValue => {
  if (dateValue) {
    return dateValue.length === 10 ? isValid(new Date(dateValue)) : dateValue.length === 0
  }
  return true
})
```

But for my next validation, I need to check that the date given is after a date in another field.  

How can I pass both values to the function to be compared?  

A specific solution would be great, but even better would be a more generalized writeup on how to make custom validators that have access to values in the data model. 

I'm not trying to take over this thread, but it seems like the right place to post about the docs being unclear on custom validators.

[...]

PR #294, opened 2018-05-21
Export validator helpers and improve documentation for custom validators

### LINKED ISSUE #193: Update documentation on custom validators
Currently a custom validators should follow ES6 and there's not much documentation or an example about it.

Here is the `maxLength` thing as an external custom validator

```js
import {req, len} from 'vuelidate/lib/validators/common'
import {withParams} from 'vuelidate'

export default length => withParams(
    { type: 'maxLength', max: length },
    value => !req(value) || len(value) <= length
)

// then where you want to use it
import maxLength from ./maxLength
```

Please add/enhance the documentation on this matter.

Frizi:
closes #193, #168, #260, #253, #240

### DIFF
--- a/docs/partials/validators/_validators.pug
+++ b/docs/partials/validators/_validators.pug
@@ -101,7 +101,7 @@ mixin validatorRow(validator, params)
             +validatorRow('and', ['validators...'])
               | Passes when all of provided validators passes.
             +validatorRow('withParams', ['$params', 'validator'])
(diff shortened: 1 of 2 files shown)


lawsonry:
Not that my opinion matters here, but I just wanted to say that I really like and appreciate the changes to the docs made here. <3

Frizi:
@lawsonry Your opinion does matter and it's really great and positive to know that people care. Thank You!

PR #316, opened 2018-06-01
Fix $anyError

### LINKED ISSUE #315: $anyError should check $error on children instead of $anyDirty
Currently $anyError uses this expression according to the docs:

```js
this.$anyDirty && !this.$pending && this.$invalid
```

This is wrong because a form can be `$invalid` and `$anyDirty` but not having any single `$error` in the validation tree.

For example, you only have required fields.
1. The form is `{ $invalid: true, $anyDirty: false, $anyError: false }` and no errors are displayed (yet)
2. The user enter something in the first field, making it valid
3. We touch the field so it becomes `$dirty`
5. No error are show yet, because the other fields aren't `$dirty` (the user hasn't come to those yet)
4. The form is now `{ $invalid: true, $anyDirty: true, $anyError: true }`

It should be `{ $invalid: true, $anyDirty: true, $anyError: false }` because we don't display any error.

This is very anoying if we want to disable the submit button *as long as* at least one error is shown.

Akryum:
Close #315

### DIFF
--- a/docs/partials/api/_values.pug
+++ b/docs/partials/api/_values.pug
@@ -60,8 +60,8 @@
           td.table__td: strong boolean
           td.table__td
             | Convenience flag to easily decide if a message should be displayed. A variant that considers
-            | error to be displayed when <strong>any</strong> of it's children are <kbd>$anyDirty</kbd>.
-            | Equivalent to <kbd>this.$anyDirty && !this.$pending && this.$invalid</kbd>.
+            | error to be displayed when itself or <strong>at least one</strong> of its children has <kbd>$error</kbd>
+            | equal to <kbd>true</kbd>.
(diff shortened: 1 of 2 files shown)


MarredCheese:
FYI the [documentation ](https://vuelidate.netlify.com/#sub-v-values) for `$anyError` currently states:

> Convenience flag to easily decide if a message should be displayed. A variant that considers error to be displayed when any of it's children are $anyDirty. Equivalent to `this.$anyDirty && !this.$pending && this.$invalid`"

So it seems commit https://github.com/vuelidate/vuelidate/pull/316/commits/8215b0df573ac1432ad0ce8a6aa2cd0571dd00c0 got trampled at some point.

PR #411, opened 2018-12-05
Added $valid value

lukaszb:
I've added `$valid` getter as I sometimes want to use it and it's easier to read than `!$v.value.$invalid` (no need for double negation).

Let me know if this makes sense and if you want me to update docs etc.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -72,6 +72,9 @@ const validationGetters = {
       this.ruleKeys.some((rule) => !proxy[rule])
     )
   },
+  $valid() {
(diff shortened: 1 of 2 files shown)


lukaszb:
@dobromir-hristov mind to explain why this has been closed?

dobromir-hristov:
Vuelidate 0.x is not going to get any more feature updates. All development focus is now on v2.x

Thank you for your contribution and time.

lukaszb:
@dobromir-hristov all right, got it, thanks!

PR #418, opened 2018-12-11
Allows Swedish characters in email domain name

JSandmark:
Under the Swedish domain .SE domains are allowed to include the three nordic characters å, ä and ö. (Capital Å, Ä and Ö)
The email validator should allow nordic domains including these characters.

Currently the email validator does not accept these characters.

Any better way to include them in the regex would be much appreciated but here is a suggestion, thanks for a great library!

NOTE: There are actually a lot more characters allowed under .se but these three are the one that is most used (https://www.iis.se/docs/teckentabell-04.pdf)


dobromir-hristov:
Not sure about this. If we add this, we need to cater for other types of emails as well. You can easily use your own email validator, as I suspect you have.
@shentao whats your take on this?

shentao:
Depending on how much hassle it is and how potentially problematic it can turn. I'd say, bring your own validator. Adding them is extremely easy after all.

dobromir-hristov:
Alright. Closing this for now. Version 2.0 may use an entirely external validation package.

Issue #419, opened 2018-12-12
Proposition : $errors as array of non passing validators


y-nk:
Hi !

I'd like to know if it would be possible to add an array of validator's names (`$errors`) for each non passing validator of each value : 

**How would it look like :**

```js
// here nothing change
validations: {
  foo: { required, minLength: minLength(4) }
}

// ... somewhere later (or by user input)

foo = ''
console.log($v.foo.$errors) // ['required, 'minLength']
foo = 'foo'
console.log($v.foo.$errors) // ['minLength']
foo = 'foobar'
console.log($v.foo.$errors) // []
```

**This would enable :** 

- validator priority (defined by key order)
- possibility to list all the errors in template : `<strong v-for="error in $v.foo.$errors">{{ error }}</strong>`
- possibility to use validator key as reference for i18n : `<strong v-if="$v.foo.$errors.length">{{ $t($v.foo.$errors[0]) }}</strong>`

**How hard would that be :** 

Probably not much more than a new validation getter such as : 

```
  $errors() {
    const proxy = this.proxy
    return this.ruleKeys.filter((rule) => !proxy[rule])
  },
```

<br>

If this feature is not already coded, and you're willing to accept it but have no time to code it, let me know and i'll submit a PR.

Thanks for reading !

thedamon:
I was looking for something similar in order to put "We found X errors with your submission" (and then possibly list and link to each field that has an error at the top of a form. This is a really useful thing to do for accessibility. I realized there is no straightforward way to access which fields have errors (complicated by nesting of course). This looks like a pretty good and flexible approach to enable that and other things.

y-nk:
@NataliaTepluhina thanks for adding the feature request tag. Does that mean I should submit a PR or are you guys still considering it?

luckylooke:
Related issues:
https://github.com/vuelidate/vuelidate/issues/149
https://github.com/vuelidate/vuelidate/issues/339
https://github.com/vuelidate/vuelidate/issues/325

All ends up with advice:
Use the https://github.com/dobromir-hristov/vuelidate-error-extractor for now.

thedamon:
The vuelidate-error-extractor doesn't seem to do what this suggests. It seems to be entirely oriented around templating. I already have my templating pattern figured out and would just like an array of error fields in the style of vuelidate (ie data only) for a11y purposes.

scniro:
[...]

Issue #149, opened 2017-05-31
Error Messages Helper Function


Austio:
Hello,

What do you think about a function like `$errors` similar to $flattenParams that collects a generic list/dictionary of errors from vuelidate parameters.

This is specific to my implementation but helper would be something slightly more generic than this that doesn't assume we want the last param in the form path..

```
          const params = this.$v.form.$flattenParams();
          const invalidParams = params.reduce((acc, param) => {
            const path = param.path.join('.');
            const validation = param.name;

            const isValid = _get(this, `$v.form[${path}][${validation}]`, true);
            if (!isValid) {
              const paramNameLocation = param.path.length - 1;

              acc.push({ [param.path[paramNameLocation]]: param.name });
            }
            return acc;
          }, []);

         return _isEmpty(invalidParams) ? false : invalidParams;

// Returns a list like this
[
  { 
    "city_id": "required"
  }
]
```

Usage would be something like this
```
submitForm() {
  if ($v.form.$errors) {
    // Handle stuff that is errored from the form 
  }
}

or in template

<p v-for="(error, index) in $v.form.$errors">
  {{ error | myErrorFormatter }}
</p> 
```

I have read through #3 and a couple other issues with discussion on passing in custom messages.  I think this is different because it is a helper for collecting data that can be formatted however someone would want.

Frizi:
Thanks for doing the hard part of figuring things out here :)

First point, i think that `$errors` can get pretty easily confused with `$error` prop.
Also, you don't have to base your implementation on `$flattenParams`really (and `_get`, that looks fragile). But it does surely work as a good proof of concept, more thoughtful implementation can be easily done later. I'd like to first think about desired API instead of focusing on implementation.

By my personal taste, I'd construct the template part more around custom components, like i previously suggested doing in https://github.com/monterail/vuelidate/issues/3#issuecomment-276457558
But that certainly isn't the only way to do it. What i'm missing in your proposal is the distinction between *local errors* and *nested errors*. Currently, local errors are just a collection props on your validation object, and for global ones you really have to traverse the tree.  Also there is no concept of order in validators currently, you kinda just grab them in random one (which might often be just fine).

Going back, flattenParams` was meant to make the second case easier, while still providing you enough information to filter out only the things you are interested in. It also gave you only the "metadata" and set of paths, with the idea that you still might easily get the actual field state using it. So it kinda avoided the problem in the first place, while introducing a second one - not being terribly useful.

[...]

PR #420, opened 2018-12-12
Feature/errors getter

y-nk:
First referred as #419 **EDIT:** also #149 (a weird anagram-ish coincidence if you ask me) 

This will enable to have a `$errors : Array` which contains all the failing rules keys. It allows to use the keys with `vue-i18n` easily and iterate over errors if necessary.

I checked on the #149 thread and honestly all the solutions proposed are overkill for a feature which so many people asked for, regarding the code which is actually implied.

Some thoughts about that thread (if you wish to move back to #149 let me know) :

- About the "you should build it with custom components" : for CMS (which are mainly using forms), a lot - if not most - of them are using ui libraries such as Vuetify, Element, and so on (even sometimes custom ones). For most implementations, error fields are props that you pass along, and not template that you can build and insert. For such reason, most of the time, this solution don't work.

- About the "$errors name is confusing" point : I can hear it, tho it seems obvious enough that all the third party proposals are basically using it `${prefix}Errors` (`$errors`, `getValidationErrors`, `$allErrors`, ... you see my point). this library already uses `$error : Boolean`, why not `$errors : Array`. If it's written in the docs, what's so non obvious about it ?

About the recursivity : i understand the need, since you can also use vuelidate in the context of a form. Nevertheless, this feature aiming at targeting directly a field, it seems totally normal not to be recursive ; although i'm sure we could easily add the recursive path as a prefix to the key (`form.email.required`, for example)

About the sorting : sure it's needed. I know it's not "obvious" and "enforced", but everyone coding javascript understand that key order matters in an object. For such reason, even if it seems "not obvious enough" or "prone to error" on an ideologic level, i think it actually is enough for javascript developers to be happy with the feature.

Finally. This demand has been commented for few months (Jun 1st to Oct 27th) last year, and it's been 2 years already, and all i can see is "thinking about the APi". I totally understand and would be more than happy to help on that matter, but also would wish to have something - even temporary - delivered, as it seems blocking for a lot of project using this more than needed library.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -97,7 +97,7 @@ const validationGetters = {
   },
   $anyError() {
     if (this.$error) return true
-    
+
     return this.nestedKeys.some((key) => this.refProxy(key).$anyError)
   },
   $pending() {
--- a/src/index.js
+++ b/src/index.js
@@ -95,6 +95,10 @@ const validationGetters = {
   $error() {
     return this.$dirty && !this.$pending && this.$invalid
   },
+  $errors() {
+    const proxy = this.proxy
+    return this.ruleKeys.filter((rule) => !proxy[rule])
+  },
   $anyError() {
     if (this.$error) return true
 


y-nk:
still no update ? please 🙏

y-nk:
well...

PR #432, opened 2019-02-03
add georgian letters in alpha & alphanumeric validators

giokaxo:
Alpha and Alphanumeric validators didn't support Georgian alphabet.
I've added Georgian letters in regex.


dobromir-hristov:
They also do not support Cyrillic. Maybe we should think of another validator? @shentao

shentao:
The change to the regex isn’t that big and I assume it doesn’t affect the performance all that much. I, however, wonder if having a new validator for that would be great since those would be colliding with each other unless you use an `or`. Which might be a bit counter-intuitive. Might be better to include all those regexes inside one validator.

dobromir-hristov:
@shentao I meant a new way to validate emails, as in a totally different regex, replacing the old one, not yet another one, just for emails with funky letters.

shentao:
Oh I see, yeah, in that case I’m all for that! :)

dobromir-hristov:
Researched a bit more on the topic. Turns out ist not easy at all to validate any kind of alphanum character. We either have to define all the possible ranges (which is insane) or use some pretty hefty libraries 😭 I am open for suggestions here.

shrpne:
To strictly validate all languages regex should be like this
https://github.com/slevithan/xregexp/blob/master/tools/output/scripts.js or this
https://github.com/slevithan/xregexp/blob/master/tools/output/blocks.js
But 20KB regex is not what everyone wants.

How about to split this into to regexes alpha which will be light Latin only and alphaUnicode for others

PR #455, opened 2019-03-15
Updated the `required` validator to strip whitespace of strings so …

cfoucher:
…that pure whitespace strings fail the validation.

- Fixed lint warnings


nolde:
Hello!

Did this ever get released? I don't see it listed in releases. We are currently using **vuelidate v0.7.4**, and required still accepts spaces as proper input.

PR #478, opened 2019-05-08
Create maxCharLength.js

afwn90cj93201nixr2e1re:


dobromir-hristov:
Could explain why you made this and how it diverges from the already existing `maxLength`? 
Also adding a tests would be necessary to accept this PR

afwn90cj93201nixr2e1re:
Emoji and other stuff - 1 emoji - 4bytes, but 1 symbol. I mean i think you get what im talking about. Im not good in tests, xD, if you can, just add it.
maxLength 4 bytes symbol return 1;
maxCharLength 4 bytes symbol return 4;

So, it's very helpfull with password's, where it's very important, coz some hash functions depends on max char length, not on length.

dobromir-hristov:
Ooh... well now heh had not thought of emoji scenarios... You sure this is the proper way to validate this? (legit question, I do not know 😆 )

As for the test, I cant push to your branch. Check out the one for maxLength and copy it, but use emojis or Chinese symbols (those count, right?) to test it out. If you still cant, I will try to help out.

afwn90cj93201nixr2e1re:
>You sure this is the proper way to validate this? (

Yeah, i tested it on backend very well on password scen. and bcrypt2.

Ok, i'll try it tomorrow.

PR #491, opened 2019-06-13
Implement validators.js

orblazer:
Hello,
I have implement [validators.js](https://github.com/chriso/validator.js) for have many default rules.

/!\ This PR change/could break rules :
-  `url`: Now this need use with `url()`
-  `integer`: Now this need use with `integer()`
-  `numeric`: Now this need use with `numeric()` and the decimal is valid
-  `decimal`: Now this need use with `decimal()`
-  `macAddress`: Now we could not use custom separator and the empty separator is `macAddress(true)`
-  `ipAddress`: Now this need use with `ipAddress()`
-  `alphaNum`: Now this need use with `alphaNum()`
-  `alpha`: Now this need use with `alpha()`

Note: The test is missing and i need help for write that.

> **BEFORE MERGE** : update docs and probably tests

Thanks.


orblazer:
Hello, some news about that PR ?

dobromir-hristov:
Hello @orblazer. Sorry for such a late reply, you were not forgotten, just lots of going back and forth.

This PR, is outstanding! However, as great as it is, it inroduces a breaking change, as most validators now require to be invoked as functions, even if they need no params. To be honest with you, you should not have started implementing it, without asking. We plan to leave the current API as is.

As stated in the roadmap, a complete rewrite is planned, where such a PR could actually be useful. I also think that implementing something like `validators.js` is better in the long run. Battle tested an all. 

I can offer you this, you are very much free to release this as an external library, `vuelidate-extra-validators` or something similar, or cooler. That way, everyone can decide if they want to upgrade and go that route. I can make sure to state its an alternative :) We do plan anyway to extract the validators away from the core, so we may add it later as a sub package or something, if you are OK with it :)

Thanks again.

orblazer:
Ok thanks, I will write a package when I have some time again and use `vuelidate` again.

PR #516, opened 2019-08-03
allow deep requiredIf string

mcshaz:
this line of code allows the requiredIf validator (and any validators using the ```ref``` helper) to use dot notation to describe deep properties ('foo.bar'). see the added tests.

### DIFF
--- a/src/validators/common.js
+++ b/src/validators/common.js
@@ -38,7 +38,7 @@ export const len = (value) => {
 export const ref = (reference, vm, parentVm) =>
   typeof reference === 'function'
     ? reference.call(vm, parentVm)
-    : reference.split('.').reduce((parent, prop) => parent[prop], parentVm);
+    : reference.split('.').reduce((parent, prop) => parent === void 0 ? void 0 : parent[prop], parentVm);
 
 // regex based validator template
 export const regex = (type, expr) =>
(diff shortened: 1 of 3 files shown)


dobromir-hristov:
This is not a bad idea. Please add tests for the `ref`, not the validator, as we are changing that, no the validator.

@shentao what do you think? I dont see it introducing anything breaking, but should allow for easier deep parent data access. Ofc this is not a perfect solution, it is kinda slow and does not allow for array index access. We could use this as its pretty fast and covers most cases - https://github.com/dobromir-hristov/get-value

mcshaz:
Tests now expanded. It would of course be a breaking change for anyone using dots in any string properties of an object - but that is a pretty edge case, and IMHO people need to learn not to choose property names like that. Can't argue about array index access. 

I would be interested in any tests you have done benchmarking the get-value library against the non-optimised but succinct function used here in real world scenarios (nested 0 to 2 properties deep in 99% of cases). 

Your guys library of course, but I chose vuelidate (over vee-validate) for its size* - as soon as you add too many fancy functions this may become less of a selling point. I thought this PR may have a place as it adds so few extra characters.

*the advantage of smaller libraries for me is designing a PWA for mobile devices, I can limit the size of the app.

Issue #343, opened 2018-07-27
minLength(1) validates true empty array


romankierzkowski:
minLength(1) must be used with required otherwise minLength(1) validates true empty array:

How to reproduce:
https://jsfiddle.net/b5v4faqf/4979/

Expected behaviour:
Setting minLenght(1) should make empty array $v.array.minLength == true.

Current behaviour:
$v.array.minLength should be false for empty array.

OCJvanDijk:
Most built-in validators return true for 'empty' values. This is so you can separately specify if a value is required or not. Maybe an e-mail is not required but once you fill one in it has to be valid. For arrays I can't immediately think of an obvious example like that but it still works the same way. So add a 'required' validator to "array" in your example and it works.

PR #541, opened 2019-09-16
Update _validators.pug

joelpro2:
Notation for minLength(1) should be present, according with the issue below, to avoid confusion.
https://github.com/vuelidate/vuelidate/issues/343


dobromir-hristov:
Hey I think that is common sense. Sorry for the late reply, just cleaning up.

joelpro2:
If there is more than 1 person confused with it, then it's probably not common sense

PR #609, opened 2020-01-04
Аdd promise as property resolver to requireIf

### LINKED ISSUE #605: requiredIf doesn't work with return promise of boolean
requiredIf doesn't seem to support the passed function returned a promise of a boolean. It looks like it considers the promise a truthy value, and thus the field is always required.

It would be great to support a promise.

dobromir-hristov:
This PR should allow resolving props in the requiredIf and requireUnless validators.

Fixes #605


dobromir-hristov:
I could not figure a more clever way to detect whether it is a function or a promise. The other way would be to always await the property, but then every `requiredIf` will be async.

PR #629, opened 2020-02-17
Unwrap later

shentao:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

**Other information:**

### DIFF
--- a/packages/vuelidate/src/core.js
+++ b/packages/vuelidate/src/core.js
@@ -1,5 +1,5 @@
-import { isFunction, isPromise, paramToRef, unwrap, unwrapObj } from './utils'
-import { computed, reactive, ref, watch, toRefs } from 'vue'
+import { isFunction, isPromise, isUndefined, objectToRefs, paramToRef, unwrap, unwrapObj } from './utils'
+import { computed, reactive, ref, watch, toRefs, isRef, getCurrentInstance } from 'vue'
 
 /**
  * @typedef NormalizedValidator
@@ -60,7 +60,7 @@ function sortValidations (validations) {
 
 function callRule (rule, value) {
   const v = unwrap(value)
-  return rule(v)
+  return rule.call(getCurrentInstance(), v)
 }
 
 /**
@@ -218,12 +218,11 @@ function createValidationResults (rules, state, key, parentKey) {
   }
 
   watch(rules, (changedRules) => {
+    const ruleKeys = Object.keys(changedRules)
     /**
      * If there are no validation rules, it is most likely
      * a top level state, aka root
      */
-    const ruleKeys = Object.keys(changedRules)
-    // if (!ruleKeys.length) return result
     ruleKeys.forEach(ruleKey => {
       results[ruleKey] = createValidatorResult(
         changedRules[ruleKey],
@@ -231,7 +230,6 @@ function createValidationResults (rules, state, key, parentKey) {
         key
       )
     })
-
     results.$invalid = computed(() =>
       ruleKeys.some(ruleKey => results[ruleKey].$invalid.value)
     )
@@ -278,7 +276,7 @@ function collectNestedValidationResults (validations, state, key) {
     collected.value = nestedValidationKeys.reduce((results, nestedKey) => {
       // if we have a key, use the nested state
       // else use top level state
-      const nestedState = key ? state[key] : state
+      const nestedState = !isUndefined(key) ? state[key] : state
       // build validation results for nested state
       results[nestedKey] = setValidations({
         validations: changedValidations[nestedKey],
@@ -312,22 +310,20 @@ function createMetaFields (results, ...otherResults) {
 
   const $errors = computed(() => {
     // current state level errors, fallback to empty array if root
-    const modelErrors = unwrap(results.$errors) || []
-
     // collect all nested and child $errors
     const nestedErrors = allResults.value
-      .filter(result => result.$errors.value.length)
+      .filter(result => result.$errors.length)
       .reduce((errors, result) => {
-        return errors.concat(...result.$errors.value)
+        return errors.concat(result.$errors)
       }, [])
 
     // merge the $errors
-    return modelErrors.concat(nestedErrors)
+    return results.$errors.value.concat(nestedErrors)
   })
 
   const $invalid = computed(() =>
     // if any of the nested values is invalid
-    allResults.value.some(r => r.$invalid.value) ||
+    allResults.value.some(r => r.$invalid) ||
     // or if the current state is invalid
     unwrap(results.$invalid) ||
     // fallback to false if is root
@@ -336,7 +332,7 @@ function createMetaFields (results, ...otherResults) {
 
   const $pending = computed(() =>
     // if any of the nested values is pending
-    allResults.value.some(r => r.$pending.value) ||
+    allResults.value.some(r => r.$pending) ||
     // if any of the current state validators is pending
     unwrap(results.$pending) ||
     // fallback to false if is root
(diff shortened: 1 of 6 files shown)


packages/vuelidate/src/core.js
```diff
@@ -23,38 +23,30 @@ import { computed, reactive, ref, watch } from 'vue'
...
-      default:
-        nestedValidators[key] = v
-    }
-  })
+  let rules = ref({})
+  let nestedValidators = ref({})
+  let config = ref({})
+
+  watch(validations, (changedValidations) => {
+    // rules.value = {}
```
shentao: Don’t we want to reset those?

packages/vuelidate/src/core.js
```diff
@@ -23,38 +23,30 @@ import { computed, reactive, ref, watch } from 'vue'
...
+        rules.value[key] = v
+      } else if (isFunction(v)) {
+        rules.value[key] = { $validator: v }
+      } else if (key.startsWith('$')) {
+        config.value[key] = v
+      } else {
+        nestedValidators.value[key] = v
+      }
+    })
+  }, { flush: 'sync', deep: true })
```
shentao: Why is the `sync` flag needed here? Wouldn’t `deep` be enough?

packages/vuelidate/src/core.js
```diff
@@ -150,19 +143,18 @@ function createAsyncResult (rule, model, initResult, $pending) {
  * @return {{$params: *, $message: Ref<String>, $pending: Ref<Boolean>, $invalid: Ref<Boolean>}}
  */
 function createValidatorResult (rule, state, key) {
-  const model = computed(() => state[key].value)
-  const ruleResult = callRule(rule.$validator, model)
+  const ruleResult = callRule(rule.$validator, state[key])
```
shentao: We can get rid of this part with the lazy validations I implemented. Do you think it could be further improved with the changes that you made now?

packages/vuelidate/src/core.js
```diff
@@ -217,58 +209,55 @@ function createValidatorResult (rule, state, key) {
...
-   * a top level state, aka root
-   */
-  if (!ruleKeys.length) return result
+  watch(rules, (changedRules) => {
+    /**
+     * If there are no validation rules, it is most likely
+     * a top level state, aka root
+     */
+    const ruleKeys = Object.keys(changedRules)
+    // if (!ruleKeys.length) return result
```
shentao: I think the above comment targets this commented line. Do you know if we can early return here? If so, it might be a good idea to skip the following operations (even though on an empty array it would be pretty straightforward. Though we would need to add the `$invalid`, `$pending` and other keys to the `results` object.

packages/vuelidate/src/core.js
```diff
@@ -217,58 +209,55 @@ function createValidatorResult (rule, state, key) {
...
+      ruleKeys.some(ruleKey => results[ruleKey].$pending.value)
+    )
+    results.$error = computed(() =>
+      results.$invalid.value && $dirty.value
+    )
+    results.$errors = computed(() => ruleKeys
+      .filter(ruleKey => results[ruleKey].$invalid.value)
+      .map(ruleKey => {
+        const res = results[ruleKey]
+        return reactive({
```
shentao: Does this need to be `reactive`? It’s already inside a computed property, so I would assume it would stay updated. Unless you use that to unwrap the `res.$message` and others?

packages/vuelidate/src/core.js
```diff
@@ -217,58 +209,55 @@ function createValidatorResult (rule, state, key) {
...
+          $validator: ruleKey,
+          $message: res.$message,
+          $params: res.$params,
+          $pending: res.$pending
+        })
       })
-    })
-  )
+    )
+  }, { flush: 'sync', deep: true })
```
shentao: Again, why `sync`?

packages/vuelidate/src/core.js
```diff
@@ -279,25 +268,29 @@ function createValidationResults (rules, state, key, parentKey) {
...
+      // build validation results for nested state
+      results[nestedKey] = setValidations({
+        validations: changedValidations[nestedKey],
+        state: nestedState,
+        key: nestedKey,
+        parentKey: key
+      })
+      return results
+    }, {})
+  }, { flush: 'sync', deep: true })
```
shentao: 😅

packages/vuelidate/src/core.js
```diff
@@ -279,25 +268,29 @@ function createValidationResults (rules, state, key, parentKey) {
...
+
+  watch(validations, (changedValidations) => {
+    const nestedValidationKeys = Object.keys(changedValidations)
+    // if we have no state, return empty object
+    if (!nestedValidationKeys.length) return
+
+    collected.value = nestedValidationKeys.reduce((results, nestedKey) => {
+      // if we have a key, use the nested state
+      // else use top level state
+      const nestedState = key ? state[key] : state
```
shentao: More of a comment to the initial implementation of this, but I’m wondering – shouldn’t we use `nestedKey` here? Given that state and validation rules should map 1:1, we should be able to use the same keys, right? Unless one of those two is behind when it comes the "depth" we are currently. Wdyt?

packages/vuelidate/src/core.js
```diff
@@ -323,9 +316,9 @@ function createMetaFields (results, ...otherResults) {
 
     // collect all nested and child $errors
     const nestedErrors = allResults.value
-      .filter(result => result.$errors.length)
+      .filter(result => result.$errors.value.length)
       .reduce((errors, result) => {
-        return errors.concat(...result.$errors)
+        return errors.concat(...result.$errors.value)
```
shentao: If `$errors` is an array, I don’t think we need to spread it to use it with concat.

packages/vuelidate/src/core.js
```diff
@@ -459,8 +451,7 @@ export function setValidations ({ validations, state, key, parentKey, childResul
     )
   }
 
-  return reactive({
-    ...results,
+  return computed(() => Object.assign({}, results, {
```
shentao: Haven’t you said that returning a computed was a bad idea?

packages/vuelidate/src/core.js
dobromir-hristov: Yes, but it was causing the entire tree to rebuild, so I was trying out things.

packages/vuelidate/src/core.js
dobromir-hristov: No, because your tree won't be built at first, so you would get errors that $v.number.required is undefined. At least that's what I got.

packages/vuelidate/src/core.js
dobromir-hristov: Correct. Have not synced yet.

packages/vuelidate/src/core.js
dobromir-hristov: We can yeah. I had it at one point, but forgot to re add it.
We fallback to false in the extractMeta method.

packages/vuelidate/src/core.js
dobromir-hristov: It's is to expand the properties, correct. Otherwise they are with refs

packages/vuelidate/src/core.js
dobromir-hristov: Nested key is the name of the validator, not the property it validates, right?

packages/vuelidate/src/core.js
dobromir-hristov: I think you are right

packages/vuelidate/src/core.js
dobromir-hristov: I cannot think of a way, to make this spreadable and reactive, at the same time.... Any ideas?

PR #662, opened 2020-05-16
Options api alternative

shentao:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
Alternative approach to handling Options API. Also adds the lazy validations to see if this approach is not triggering recalculations where it shouldnt.


packages/vuelidate/src/core.js
```diff
@@ -147,20 +148,19 @@ function createAsyncResult (rule, model, initResult, $pending) {
  * @param {String} key
  * @return {{$params: *, $message: Ref<String>, $pending: Ref<Boolean>, $invalid: Ref<Boolean>}}
  */
-function createValidatorResult (rule, state, key) {
-  const model = computed(() => state[key].value)
-  const ruleResult = callRule(rule.$validator, model)
+function createValidatorResult (rule, state, key, $dirty) {
+  const model = computed(() => unwrap(unwrap(state)[key]))
```
dobromir-hristov, line 152: whats with all the unwrapping?

packages/vuelidate/src/index.js
```diff
@@ -64,12 +65,14 @@ export const VuelidateMixin = {
       : options.validations
 
     if (!options.computed) options.computed = {}
-    if (options.computed.$v) return
+    if (options.computed.v$) return
```
dobromir-hristov, line 68: is `$v` still not allowed? Didnt Evan change it?
I think this will very much confuse people :/

packages/vuelidate/src/core.js
```diff
@@ -91,8 +91,9 @@ function normalizeValidatorResponse (result) {
  * @param {Ref<*>} model
  * @return {Ref<Boolean>}
  */
-function createComputedResult (rule, model) {
+function createComputedResult (rule, model, $dirty) {
```
dobromir-hristov, line 94: This is part of the lazy by default functionality, right?

packages/vuelidate/src/core.js
```diff
@@ -147,20 +148,19 @@ function createAsyncResult (rule, model, initResult, $pending) {
...
-function createValidatorResult (rule, state, key) {
-  const model = computed(() => state[key].value)
-  const ruleResult = callRule(rule.$validator, model)
+function createValidatorResult (rule, state, key, $dirty) {
+  const model = computed(() => unwrap(unwrap(state)[key]))
 
   const $pending = ref(false)
   const $params = rule.$params || {}
-  const $invalid = isPromise(ruleResult)
+  const $invalid = rule.$async
```
dobromir-hristov, line 156: Is this the new convention where we ask them to apply this to async validators?

What do we do with backwards compat? We need to have some sort of fallback or detection, otherwise the validator will always return truthy, the `promise` is an object :)

Maybe we can check if after calling this in the computed, it returns a promise, and if it does, we throw or warn?

packages/vuelidate/src/core.js
```diff
@@ -449,10 +450,9 @@ export function setValidations ({ validations, state, key, parentKey, childResul
   }) : null
 
   if (config.$autoDirty) {
```
dobromir-hristov, line 452: Have we tested if this scenario works?

```js
validations: () => ({
  name: { required },
  surname: { anything },
  $autoDirty: true
})
```

packages/vuelidate/src/core.js
```diff
@@ -106,31 +107,31 @@ function createComputedResult (rule, model) {
...
+      $pendingCounter.value++
+      $pending.value = !!$pendingCounter.value
       $invalid.value = true
 
       ruleResult
         .then(data => {
-          $pending.value = false
+          $pendingCounter.value--
+          $pending.value = !!$pendingCounter.value
           $invalid.value = !data
```
dobromir-hristov, line 130:
```suggestion
          $invalid.value = normalizeValidatorResponse(data)
```

packages/vuelidate/src/core.js
```diff
@@ -106,31 +107,31 @@ function createComputedResult (rule, model) {
  * @param {Ref<Boolean>} $pending
  * @return {Ref<Boolean>}
  */
-function createAsyncResult (rule, model, initResult, $pending) {
-  const $invalid = ref(true)
+function createAsyncResult (rule, model, $pending, $dirty) {
+  const $invalid = ref(!!$dirty.value)
+  const $pendingCounter = ref(0)
```
dobromir-hristov, line 112: Just wondering what happens, if you drop the counter? Is it to deal with racing conditions?

packages/test-project/src/App.vue
```diff
@@ -1,16 +1,19 @@
 <template>
   <div id="app">
+    <OldApiExample />
```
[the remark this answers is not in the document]
shentao: Let’s start using Vitepress for that maybe? You can see https://github.com/vuelidate/formvuelatte/tree/next for reference. It has it’s problems, but generally works OK. And is amazing for development since the updates are so fast.

packages/vuelidate/src/core.js
shentao: Yup. Sorry, I mixed the two PRs.

packages/vuelidate/src/core.js
shentao: It’s to prevent setting `$pending` to `false`, when there were multiple requests done and one already returned (and for example was `$invalid === false`), thus giving a false positive, since the later requests could return `$invalid === true`. So what happens here is just making sure that `$pending` is always `true` where there is at least 1 request still going on.

packages/vuelidate/src/core.js
shentao: Oh damn. I think it might be unnecessary!

packages/vuelidate/src/core.js
shentao: Yeah! I figured that now that we don’t do validations upfront until the state is dirty, we have no easy way to assign different handlers for validators. To keep it consistent, I figured having a `$async` flag makes it really easy to navigate the different strategy. I think that depending on user feedback we could then also implement some other ways to handle async validations. For example maybe validations based on status codes? Something to think about.

But I agree with the fallback detection. Throwing an error for the developer there would make perfect sense!

packages/vuelidate/src/core.js
shentao: Good question. I think I have not tested this or I don’t remember. But this is also something I was considering. Or have the 3rd argument in `useVuelidate` by the config object. Wdyt? Having those next to actual rules doesn’t really make that much sense.

packages/vuelidate/src/index.js
shentao: It didn’t work for me when I tried. :(

dobromir-hristov: This will cause much confusion... not to mention backwards compatibility. A search and replace is easy, but still.

dobromir-hristov: How do Vuex and VueRouter do it now?

packages/vuelidate/src/core.js
dobromir-hristov: Could you try to remove it and see if it works?

We will ofc cover this in tests later on.

packages/test-project/src/App.vue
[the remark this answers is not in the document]
dobromir-hristov: I have -1 experience with it. But I can try. This means I can drop the whole Vue CLI for the example app.

packages/vuelidate/src/core.js
dobromir-hristov: I was also thinking if we could make all validators `async` by default. This will complicate it though...
But we can ofc just throw. Should we do this here, or file a separate issue to track it later? 

Also we can offer a helper, like Vue does for async functional components. Similar to `withMessage` and `withProps`, `email: { isTaken: withAsync(asyncCheck) }`

packages/vuelidate/src/core.js
dobromir-hristov: Nope.. I dont like it either.... But what about options API? Dont treat it as a second class citizen mate :D

packages/vuelidate/src/core.js
dobromir-hristov: If we could cancel the previous request before firing the new one, that would be pretty dope :/ This is something that RX does pretty nicely too.

How would a userland cancellation look like?

```js
const cancelToken = axio.createCancelToken()

async asyncValidator() {
  cancelToken.cancel()
 try {
     const response =  await axios('place', { cancel: cancelToken.token })
  } catch(e) {
     // could also handle the cancel token or smthn
     return false
  }
  return response.isValid
}
```

shentao: Let’s introduce that in another PR to unblock this one.

packages/vuelidate/src/core.js
shentao: Let’s have it as a separate PR.

packages/vuelidate/src/core.js
shentao: Good idea, but for another PR.

packages/vuelidate/src/index.js
shentao: `$v` works when added using old Options API! So it would only affect people that want to use the new `setup` approach.

packages/vuelidate/src/core.js
shentao: Unwrapped and it works nicely! Thanks!

packages/vuelidate/src/index.js
dobromir-hristov: That doesnt sound great... Its going to lead to allot of confusion.

PR #663, opened 2020-05-21
Adds validate function and other improvements

dobromir-hristov:
* Adds the `validate` method we talked about. It does not really invoke every validator, rather waits for all of them to resolve, to give you a green light.
* Adds a helper for writing `async` validators. The new syntax is not entirely short, so this should take care of that.
* Warns users if `$async` is missing from async validators

- [x] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No


packages/validators/src/utils/common.js
```diff
@@ -44,3 +44,9 @@ export function isTruthy (prop) {
 export function isPromise (object) {
   return isObject(object) && isFunction(object.then)
 }
+
+export function withAsync (validator) {
+  const normalized = normalizeValidatorObject(validator)
+  normalized.$async = true
```
dobromir-hristov, line 50: This mutates the passed object, which bothers me a bit :/

packages/vuelidate/src/core.js
```diff
@@ -94,7 +94,11 @@ function normalizeValidatorResponse (result) {
 function createComputedResult (rule, model, $dirty) {
   return computed(() => {
     if (!$dirty.value) return false
-    const result = callRule(rule, model)
+    let result = callRule(rule, model)
+    if (isPromise(result)) {
+      result = false
+      console.error('Async validators must have the $async property')
```
dobromir-hristov, line 100: Think this warning will suffice?

packages/vuelidate/src/core.js
```diff
@@ -134,7 +138,8 @@ function createAsyncResult (rule, model, $pending, $dirty) {
           $pending.value = !!$pendingCounter.value
           $invalid.value = true
         })
-    }
+    },
+    { flush: 'sync' }
```
dobromir-hristov, line 142: This was necessary otherwise it calls it on next tick, thus the `$validate` method fails to trigger properly. I am not sure of the complications of this, so I might just await a tick in `$validate` instead?

packages/validators/src/utils/common.js
shentao: Yup! I had some issues with that in other helpers as it turned out to mutate built in validators... We might want to do a spread merge.

packages/vuelidate/src/core.js
shentao: Might be better to just throw an actual error. That’s what they are for.

packages/vuelidate/src/core.js
shentao: Good question. It shouldn’t require a synced update to work properly. What exactly failed without it?

packages/vuelidate/src/core.js
```diff
@@ -455,6 +460,18 @@ export function setValidations ({ validations, state, key, parentKey, childResul
     })
   }
 
+  let $validate = function $validate () {
+    return new Promise((resolve) => {
+      if (!$dirty.value) $touch()
+      // return whether it is valid or not
+      if (!$pending.value) return resolve(!$error.value)
+      const unwatch = watch($pending, () => {
```
shentao, line 468: Smart!

packages/validators/src/utils/common.js
dobromir-hristov: I will reflect this in a new PR.

packages/vuelidate/src/core.js
dobromir-hristov: This will be such a breaking change :(

packages/vuelidate/src/core.js
dobromir-hristov: Well, `$validate` triggers dirty, and then checks what has errors, before it resolves. But watchers are async, and flush after at tick. Which means for `$validate` to work, we would need to await those to finish, and then check for error states.

PR #664, opened 2020-05-22
feat: replace vue-cli with vite

dobromir-hristov:
This replaces the vue-cli test-app with Vite. We dont gain much, but its a nice start.


dobromir-hristov:
I think next steps are to add a simple vue-router so we can have a nice app to do e2e tests against.

shentao:
Looks good! I was thinking about VitePress but it’s still not that stable when it comes to production build (https://github.com/vuejs/vitepress/issues/12).

dobromir-hristov:
For docs maybe, but not e2e app

shentao:
Is there anything wrong with running e2e on the docs examples though?

dobromir-hristov:
I dont see a reason why. We can have much more detailed test cases, without making them user friendly

shentao:
True that!

PR #672, opened 2020-06-20
Next docs refactor

marina-mosti:
This PR builds upon the next documentation to provide a more guide style approach as suggested by @dobromir-hristov in #667

### DIFF
--- a/packages/docs/src/README.md
+++ b/packages/docs/src/README.md
@@ -8,7 +8,7 @@ _If you are looking for Vuelidate that supports Vue 2, use [Vuelidate 1](https:/
 
 ## Installation
 
-Installing Vuelidate is easy, and can be done with your package manager of choice.
+Installing Vuelidate is straightforward, and can be done with your package manager of choice.
 
 ```bash
 npm install @vuelidate/core @vuelidate/validators --save
--- a/packages/docs/src/README.md
+++ b/packages/docs/src/README.md
@@ -93,7 +93,7 @@ An example of our `name` property being in an error state due to it being requir
 {
   "$property": "name",
   "$validator": "required",
-  "$message": "The value is",
+  "$message": "The value is required",
   [...]
 }
 ```
--- a/packages/docs/src/README.md
+++ b/packages/docs/src/README.md
@@ -8,15 +8,19 @@ _If you are looking for Vuelidate that supports Vue 2, use [Vuelidate 1](https:/
 
 ## Installation
 
-You can install via npm
+Installing Vuelidate is easy, and can be done with your package manager of choice.
 
 ```bash
 npm install @vuelidate/core @vuelidate/validators --save
+
+// OR
+
+yarn add @vuelidate/core @vuelidate/validators
 ```
 
 ## Getting Started
 
-You can import the library and use it as a Vue plugin to enable the functionality globally on all components containing validation configuration.
+You can then import Vuelidate and use it as a Vue plugin to enable the functionality globally on all components containing validation configuration.
 
 ```js
 import Vue from 'vue'
@@ -26,7 +30,17 @@ const app = Vue.createApp(App)
 app.use(Vuelidate)
 ```
 
-Then in your component you can create a `validations` method and define your validation rules.
+Now that Vuelidate is registered as a global plugin, create a `validations` method inside your component and define your validation rules.
+
+First, import the validators that you want to use from `@vuelidate/validators`.
+
+```js
+import { required } from '@vuelidate/validators'
+```
+
+Next, add the `validations` method, it should return an object that matches the structure of your state. In this example, we have a field `name`, which is declared inside of the component's `data()`.
+
+Notice that in the `validations` method we declare the same structure, and the `name` property specifies the validators that we want to apply to it - in this case, a `required` validation.
 
 ```js
 import { required } from '@vuelidate/validators'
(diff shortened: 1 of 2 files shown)


packages/docs/src/README.md
```diff
@@ -26,7 +30,17 @@ const app = Vue.createApp(App)
...
-Then in your component you can create a `validations` method and define your validation rules.
+Now that Vuelidate is registered as a global plugin, create a `validations` method inside your component and define your validation rules.
+
+First, import the validators that you want to use from `@vuelidate/validators`.
+
+```js
+import { required } from '@vuelidate/validators'
+```
+
+Next, add the `validations` method, it should return an object that matches the structure of your state. In this example, we have a field `name`, which is declared inside of the component's `data()`.
```
shentao, line 41: I don’t think we should call `validations` a method. It can be a function, but can also be just an object (when you don’t need `this` like in this example. Also, using the name `method` suggest adding a `validations` method to `methods`.

marina-mosti:
> This is very nice, thanks @marina-mosti! I’m wondering if we should add more in-depth explanation of how `$model` works? Mentioning that it is a getter/setter combo? Dunno if it’s actually necessary?

@shentao that's not a bad idea, maybe not super in depth but mentioning it. I'll give it a look later

packages/docs/src/README.md
```diff
@@ -26,7 +30,17 @@ const app = Vue.createApp(App)
...
+
+Notice that in the `validations` method we declare the same structure, and the `name` property specifies the validators that we want to apply to it - in this case, a `required` validation.
```
shentao, line 43: `in the validations method we declare the same structure` this is slightly unclear to me. 

I think this would be best explained visually.
1) Show a combination of data (ideally with an object inside with their own properties) + a computed property
2) Show how the above local state is then available on `this`.
3) Show how that structure maps 1:1 to what we return from `validations`.

packages/docs/src/guide.md
```diff
@@ -1,10 +1,13 @@
 # Guide
 
 ## Basics
-As we mentioned before, validation rules are set inside an object, that is returned by the `validations` method. We will refer to those as just _validations_.
+As we mentioned in the [getting started quick guide](/#getting-started-2), validation rules are set inside an object, that is returned by the `validations` method. We will refer to those as just _validations_ from now on.
```
shentao, line 4:
```suggestion
As we mentioned in the [getting started quick guide](/#getting-started-2), validation rules are defined inside the object returned by the `validations` method. We will refer to those as just _validations_ from now on.
```

packages/docs/src/guide.md
```diff
@@ -29,16 +32,20 @@ export default {
...
 
-Vuelidate builds a validation state, that can be accessed via the `v$` property. It is a nested object that roughly follows your `validations` structure,
-but with some extra validity related properties. It is reactive, meaning it changes as you type.
+Vuelidate builds a validation state, that can be accessed via the `v$` property. It is a nested object that follows your `validations` structure,
+but with some extra validity related properties.
 
-So for our form above, to check whether name is valid, we would can see if the `name.$error` property is `true` or not:
+`v$` is also reactive - meaning it changes as the user types.
+
+Building up on our form example above, to check whether name is valid we can now check to see if the `name.$error` property is `true` or `false`, and display an error for our users.
```
shentao, line 48: I think we have omitted the fact that now all validators are lazy by default. This means that even if someone types anything, the `$invalid` and `$error` state will still be `false` because the validators won’t even run until a property is `$dirty.`

I imagine this might cause confusion as the examples presented just won’t work.

packages/docs/src/README.md
```diff
@@ -60,15 +60,28 @@ export default {
...
-To check in more detail if the `required` property was the one that had the problem, and display an error for our user, we will check the `required` property inside of `v$.name`
+Each object inside the `$errors` array will contain a few properties that allows us to dynamically build our error message.
+
+An example of our `name` property being in an error state due to it being required would be:
+
+```js
+{
+  "$property": "name",
+  "$validator": "required",
+  "$message": "The value is",
```
shentao, line 73: I think this is missing the "required" word.

marina-mosti: Oops, haha

packages/docs/src/guide.md
marina-mosti: Can you provide an example of how to correctly handle this? Are we suggesting they check for both $dirty and $error? Will this be addressed if we check for $errors.length?

shentao: In the previous vuelidate version all validations happened instantly and the `$invalid` state was always up to date. You then had to call `$touch` to change the `$dirty` state. `$error` was just a logical `$invalid && $dirty` right?

Now, this still works similarly for `$error` and `$errors`. The difference is that all validators are lazy by default and are not executed unless the `$dirty` state is changed to `true`. Thus until someone uses `$touch()` or `$model` or passes `$autoDirty: true`, there is no validation being done at all. @dobromir-hristov also recently introduced:
```
this.v$.$validate().then((result) => {
  console.log('Result is', result)
})
```
Which internally triggers `$touch()` and sets `$dirty` state to true. So there are 4 ways to start validating.

This kind of brings the questions – should we even have `$invalid` at this point for any other reason that backwards compatibility? That’s because once you call `$reset()` to reset the `$dirty` state, the validators will again stop being executed.

PR #673, opened 2020-06-20
Refactor validation results storage

shentao:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/packages/test-project/src/components/DynamicValidations.vue
+++ b/packages/test-project/src/components/DynamicValidations.vue
@@ -1,0 +1,110 @@
+<template>
+  <div class="SimpleForm">
+    <div class="">
+      <label>x</label>
+      <input
+        v-model="x"
+        type="number"
+      >
+    </div>
+    <div class="">
+      <label>y</label>
+      <input
+        v-model="v$.y.$model"
+        type="text"
+      >
+    </div>
+    <div class="">
+      <label>a</label>
+      <input
+        v-model="a"
+        type="number"
+      >
+    </div>
+    <div class="">
+      <label>y</label>
+      <input
+        v-model="$v.b.$model"
+        type="text"
+      >
+    </div>
+    <button @click="$v.$touch">
+      touch
+    </button>
+    <div style="background: rgba(53, 219, 53, 0.62); color: #ff9090; padding: 10px 15px">
+      <p
+        v-for="(error, index) of $v.$errors"
+        :key="index"
+        style="padding: 0; margin: 5px 0"
+      >
+        {{ error.$message }}
+      </p>
+    </div>
+    <div style="background: rgba(219, 53, 53, 0.62); color: #ff9090; padding: 10px 15px">
+      <p
+        v-for="(error, index) of v$.$errors"
+        :key="index"
+        style="padding: 0; margin: 5px 0"
+      >
+        {{ error.$message }}
+      </p>
+    </div>
+    <pre>{{ v$ }}</pre>
+    <pre>{{ $v }}</pre>
+  </div>
+</template>
+
+<script>
+import { ref, computed } from 'vue'
+import { VuelidateMixin, useVuelidate } from '@vuelidate/core'
+import { minLength, required } from '@vuelidate/validators'
+
+export default {
+  name: 'SimpleForm',
+  mixins: [VuelidateMixin],
+  setup () {
+    const x = ref(5)
+    const y = ref('a')
+
+    const rules = computed(() => {
+      return x.value >= 6
+        ? { y: { minLength: minLength(x) } }
+        : { y: { minLength: minLength(3), required } }
+    })
+
+    const v$ = useVuelidate(
+      rules,
+      { x, y }
+    )
+    return { x, y, v$, $a: 5 }
+  },
+  data () {
+    return {
+      a: 1,
+      b: 2
+    }
+  },
+  computed: {
+    c () {
+      return 5
+    }
+  },
+  validations () {
+    if (this.a > 5) {
+      return {
+        b: {
+          minLength: minLength(this.a),
+          $autoDirty: true
+        }
+      }
+    } else {
+      return {
+        b: {
+          minLength: minLength(2),
+          $autoDirty: true
+        }
+      }
+    }
+  }
+}
+</script>
--- a/packages/validators/src/utils/common.js
+++ b/packages/validators/src/utils/common.js
@@ -1,4 +1,4 @@
-import { isRef } from 'vue'
+import { unref } from 'vue'
 
 export function isFunction (val) {
   return typeof val === 'function'
@@ -14,7 +14,7 @@ export function isObject (o) {
  * @return {*}
  */
 export function unwrap (val) {
-  return isRef(val) ? val.value : val
+  return unref(val)
 }
 
 /**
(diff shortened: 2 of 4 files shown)


shentao:
Still quite WIP though. The code needs some major refactoring as its complexity is getting out of hand.

packages/test-project/src/components/DynamicValidations.vue
```diff
@@ -0,0 +1,110 @@
...
+    </div>
+    <div class="">
+      <label>a</label>
+      <input
+        v-model="a"
+        type="number"
+      >
+    </div>
+    <div class="">
+      <label>y</label>
```
marina-mosti, line 25:
```suggestion
      <label>b</label>
```

packages/validators/src/utils/common.js
```diff
@@ -14,7 +14,7 @@ export function isObject (o) {
  * @return {*}
  */
 export function unwrap (val) {
-  return isRef(val) ? val.value : val
+  return unref(val)
 }
```
marina-mosti, line 18: Are we keeping this for any particular reason? Should we refactor everything with `unref` now that its available?

packages/vuelidate/src/core.js
```diff
@@ -285,7 +294,7 @@ function createValidationResults (rules, state, key, parentKey) {
  * @param {String} [key] - Parent level state key
  * @return {{}}
```
marina-mosti, line 295: Missing some jsdocs

```suggestion
 * @param {String} [key] - Parent level state key
 * @param {*} - Path
 * @param {*} - resultsStorage
 * @return {{}}
```

packages/vuelidate/src/core.js
```diff
@@ -419,18 +429,26 @@ function createMetaFields (results, ...otherResults) {
...
  * @return {UnwrapRef<VuelidateState>}
  */
-export function setValidations ({ validations, state, key, parentKey, childResults }) {
+export function setValidations ({
+  validations,
+  state,
+  key,
+  parentKey,
+  childResults,
+  resultsStorage
```
marina-mosti, line 438: Missing jsdocs here

packages/validators/src/utils/common.js
shentao: Yeah, most likely, just haven’t had the time for it yet!

packages/test-project/src/components/DynamicValidations.vue
```diff
@@ -0,0 +1,110 @@
...
+import { minLength, required } from '@vuelidate/validators'
+
+export default {
+  name: 'SimpleForm',
+  mixins: [VuelidateMixin],
+  setup () {
+    const x = ref(5)
+    const y = ref('a')
+
+    const rules = computed(() => {
```
dobromir-hristov, line 69: @shentao do you think it would work if a specific property has a dynamic rule set?

```js
{
  y: computed(() => { x.value >= 6 ? { required } : { somethingElse } })
  x: { required }
}
```

packages/test-project/src/components/DynamicValidations.vue
```diff
@@ -0,0 +1,110 @@
...
+      return x.value >= 6
+        ? { y: { minLength: minLength(x) } }
+        : { y: { minLength: minLength(3), required } }
+    })
+
+    const v$ = useVuelidate(
+      rules,
+      { x, y }
+    )
+    return { x, y, v$, $a: 5 }
```
dobromir-hristov, line 79: `$a` is just forgotten from tests?

packages/vuelidate/src/core.js
```diff
@@ -283,9 +292,11 @@ function createValidationResults (rules, state, key, parentKey) {
  * @param {Object<NormalizedValidator|Function>} validations - The validation
  * @param {Object} state - Parent state
  * @param {String} [key] - Parent level state key
+ * @param {*} - Path
+ * @param {*} - resultsStorage
```
dobromir-hristov, line 296:
```suggestion
 * @param {String} - path
 * @param {Map} - resultsStorage
```

packages/vuelidate/src/index.js
```diff
@@ -12,8 +12,8 @@ const VuelidateSymbol = Symbol('vuelidate')
  * @param {String} registerAs
  * @return {UnwrapRef<*>}
  */
-export default function useVuelidate (validationsArg, state, registerAs) {
-  const validations = unwrap(validationsArg)
+export function useVuelidate (validations, state, registerAs) {
+  const resultsStorage = new Map()
```
dobromir-hristov, line 16: maybe `resultsCache` ?

packages/vuelidate/src/index.js
```diff
@@ -29,21 +29,22 @@ export default function useVuelidate (validationsArg, state, registerAs) {
...
+  }))
 
   if (registerAs) {
     injectToParent(validationResults, registerAs)
   }
 
   if (registerAs && childResultsKeys.value.length) {
-    return reactive({
-      ...validationResults,
+    return computed(() => ({
```
dobromir-hristov, line 44: Wont this mean that instead of unwrapping with `reactive`, we are now returning a computed with `value` instead?

dobromir-hristov: I saw you are handling it in the Mixin, but what about normal composition api usage?

packages/test-project/src/components/DynamicValidations.vue
shentao: Interesting! The number of possibilities here is getting quite overwhelming. I think it might break the current implementation. We either try to accommodate for that or we just recommend that people make the whole rules object a computed if they want to have dynamic rules? WDYT?

packages/test-project/src/components/DynamicValidations.vue
shentao: Yes. Wanted to make sure that we can now return `$` prefixed properties from setup. We can. 🎉 (No need to force people to use `v$`). But this still has some issues I found when working with the formvuelatte-vuelidate-plugin. Turns out it throws a serious error that literally freezes the browser when I try to pass `$v` as a prop in a higher order component.

packages/vuelidate/src/index.js
shentao: Good call!

packages/vuelidate/src/index.js
shentao: Oh yeah. In `setup` it then requires us to use `$v.value`. We might want to reiterate on that. The `computed` here was required so that we rerun the validations with new rules if the rules have changed. But I assume we could possibly use a combination of `reactive` and `watch` (where we specifically point at the rules as the watched ref). This could be even more efficient and remove that `.value`.

packages/test-project/src/components/DynamicValidations.vue
dobromir-hristov: I think we should try to see if we can support this with minimal changes. If not, we recommend that they make the entire validation object dynamic. 

Depends on the case really. Composition API gives too much freedom 😆

packages/test-project/src/components/DynamicValidations.vue
dobromir-hristov: Then its something for vue core to checkout. If we can, we should stick with the previous convention.

packages/test-project/src/components/DynamicValidations.vue
shentao: Thats exactly how I feel about it. We need to review the whole flow again and see where we can normalize early so that we remove some conditionals later on.

PR #674, opened 2020-06-24
fix: chain of state refs

javiertury:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Description:**

~~1. Using the mixin, I found out that `validations` was passed to `setValidations` as a reference. It was problematic. Suppose I wrote:~~
```js
defineComponent({
  validations: {
    foo: {
      required
    }
  }
});
```
~~Vuelidate interpreted it without unwrapping it first, as if I had written:~~
```js
defineComponent({
  validations: {
    value: {
      foo: {
        required
      }
    }
  }
});
```

~~This patch unwraps `validations`.~~

This first part is already fixed by @shentao in 5725a38da12848fc699c719dafa06706107f0374

2. Refs of nested objects lose reactivity after the first level. Imagine I can only use a ref.
```js
defineComponent({
  setup (props) {
    // "props" is a shallow reactive object, so it means "foo" is not reactive
    // All I can do is to extract a reference.
    const {
      foo
    } = toRefs(props);

    const validation = useVuelidate({
      foo: {
        bar: {
          x: {
            required
          }
        }
      }
    }, { foo });

    return {
      validation
    };
  }
});
```

Vuelidate will lose reactivity after the `foo` level. To see why, let's take a look at how vuelidate transverses the state tree.
```
    const nestedState = key ? state[key] : state
```

This translates to a simple `state['foo']['bar']['x']`. It works well for reactive objects but not for references. Even if references were unwrapped, reactivity is lost.

To preserve reactivity, I thought about chaining references using `computed`. This way references are unwrapped inside computed functions and thus preserve reactivity.
```
    const nestedState = !key ? state : computed(() => unwrap(unwrap(state)[key]))
```

There are many `unwraps` but I believe they are necessary. In the first level, the key `foo` in the state `{ foo }` is a reference. But in subsequent levels, the state itself is a reference because it was generated from `computed()`. So we have to beware of references in both `state` and `state[key]`.

**Testing:**
Right now I'm unable to run tests. it goes on a never ending loop about missing individual babel plugins starting with `@babel/plugin-transform-runtime`.

This was only tested on a personal project I am migrating from Vue 2 to Vue 3. However I plan to

### DIFF
--- a/packages/vuelidate/src/core.js
+++ b/packages/vuelidate/src/core.js
@@ -1,5 +1,5 @@
 import { isFunction, isPromise, unwrap, unwrapObj } from './utils'
-import { computed, reactive, ref, watch, isRef } from 'vue'
+import { computed, reactive, ref, watch, isRef } from 'vue-demi'
 
 /**
  * @typedef NormalizedValidator
--- a/packages/vuelidate/src/core.js
+++ b/packages/vuelidate/src/core.js
@@ -1,5 +1,5 @@
 import { isFunction, isPromise, unwrap, unwrapObj } from './utils'
-import { computed, reactive, ref, watch, isRef } from 'vue-demi'
+import { computed, reactive, ref, watch, isRef } from 'vue'
 
 /**
  * @typedef NormalizedValidator
@@ -307,7 +307,7 @@ function collectNestedValidationResults (validations, state, key, path, resultsC
   return nestedValidationKeys.reduce((results, nestedKey) => {
     // if we have a key, use the nested state
     // else use top level state
-    const nestedState = key ? state[key] : state
+    const nestedState = key ? computed(() => unwrap(unwrap(state)[key])) : state
 
     // build validation results for nested state
     results[nestedKey] = setValidations({
@@ -445,6 +445,7 @@ export function setValidations ({
   resultsCache
 }) {
   const path = parentKey ? `${parentKey}.${key}` : key
+
   // Sort out the validation object into:
   // – rules = validators for current state tree fragment
   // — nestedValidators = nested state fragments keys that might contain more validators
@@ -476,22 +477,25 @@ export function setValidations ({
    */
 
   const $model = key ? computed({
-    get: () => unwrap(state[key]),
+    get: () => unwrap(unwrap(state)[key]),
     set: val => {
       $dirty.value = true
-      if (isRef(state[key])) {
-        state[key].value = val
+      const unwrappedState = unwrap(state)
+
+      if (isRef(unwrappedState[key])) {
+        unwrappedState[key].value = val
       } else {
-        state[key] = val
+        unwrappedState[key] = val
       }
     }
   }) : null
 
   if (config.$autoDirty) {
-    const watchTarget = isRef(state[key]) ? state[key] : computed(() => unwrap(state)[key])
-    watch(watchTarget, () => {
-      if (!$dirty.value) $touch()
-    })
+    watch(
+      () => unwrap(unwrap(state)[key]),
+      () => {
+        if (!$dirty.value) $touch()
+      })
   }
 
   function $validate () {
--- a/packages/vuelidate/test/unit/specs/validation.spec.js
+++ b/packages/vuelidate/test/unit/specs/validation.spec.js
@@ -1,13 +1,14 @@
 import { computed, ref } from 'vue-demi'
 import { flushPromises, mount } from '@vue/test-utils'
-import { isEven, isOdd } from '../validators.fixture'
+import { isEven } from '../validators.fixture'
 import {
   asyncValidation,
   computedValidationsObjectWithReactive,
   computedValidationsObjectWithRefs,
   nestedComponentValidation,
   nestedReactiveObjectValidation,
-  simpleValidation
+  simpleValidation,
+  nestedRefObjectValidation
 } from '../validations.fixture'
 import { createSimpleWrapper, shouldBePristineValidationObj, shouldBeInvalidValidationObject } from '../utils'
 import { withAsync } from '@vuelidate/validators/src/common'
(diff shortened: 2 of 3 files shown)


dobromir-hristov:
Hey, looks good. I know, there is allot of unwrapping, but sometimes its needed. I was worried about deep ref usage and reactive too...

@shentao made some changes recently that also improve some parts, we will have to see if this doesnt collide with them.

In the meantime, can you give me a concrete example I can test this with? The lack of tests is on our behalf, we are in the process of writing a test suite, we've been busy with other life priorities lately :)

javiertury:
On this branch, go to `./packages/vuelidate' and rebuild it. I often forget about this obvious step and it drives me nuts.

```bash
cd ./packages/vuelidate
yarn run build
cd ../..
```

Then open a `node` terminal and setup the following environment

```js
var { reactive, ref } = require('vue');
var { required } = require('./packages/validators');
var { default: useVuelidate } = require('./packages/vuelidate');

var $autoDirty = true;

var form = ref({
  foo: {
    bar: ''
  }
});

var validationArgs = {
  form: {
    required,
    foo: {
      required,
      bar: {
        required,
        $autoDirty
      },
    }
  }
};

var validation = useVuelidate(validationArgs, { form });
```

Now trigger a change on a deeply nested property

```
> form.value = { foo: { bar: 'bar' } }
```

It can be verified that `validation` is deeply reactive

```
> validation.form.$model
Proxy [
  { foo: { bar: 'bar' } },
  {
    get: [Function: get],
    set: [Function: set],
    deleteProperty: [Function: deleteProperty],
    has: [Function: has],
    ownKeys: [Function: ownKeys]
  }
]
> validation.form.foo.$model
Proxy [
  { bar: 'bar' },
  {
    get: [Function: get],
    set: [Function: set],
    deleteProperty: [Function: deleteProperty],
    has: [Function: has],
    ownKeys: [Function: ownKeys]
  }
]
> validation.form.foo.bar.$model
'bar'
```

javiertury:
On this branch, rebuild `vuelidate`. Often I forget about this obvious step and it drives me nuts.

```bash
cd ./packages/vuelidate
yarn run build
cd ../..
```

Then setup the following environment. Notice that `form` is a ref with a deeply nested structure.

```js
var { reactive, ref } = require('vue');
var { required } = require('./packages/validators');
var { default: useVuelidate } = require('./packages/vuelidate');

var $autoDirty = true;

var form = ref({
  foo: {
    bar: ''
  }
});

var validationArgs = {
  form: {
    required,
    foo: {
      required,
      bar: {
        required,
        $autoDirty
      },
    }
  }
};

var validation = useVuelidate(validationArgs, { form });
```

Now trigger a change on a deeply nested property

```
> form.value = { foo: { bar: 'bar' } }
```

It can be verified that `validation` is deeply reactive

```
> validation.form.$model
Proxy [
  { foo: { bar: 'bar' } },
  {
    get: [Function: get],
    set: [Function: set],
    deleteProperty: [Function: deleteProperty],
    has: [Function: has],
    ownKeys: [Function: ownKeys]
  }
]
> validation.form.foo.$model
Proxy [
  { bar: 'bar' },
  {
    get: [Function: get],
    set: [Function: set],
    deleteProperty: [Function: deleteProperty],
    has: [Function: has],
    ownKeys: [Function: ownKeys]
  }
]
> validation.form.foo.bar.$model
'bar'
```

javiertury:
- @shentao fixed in 5725a38da12848fc699c719dafa06706107f0374 the issue about unwrapped `validations`. I removed some code since it's now redundant.
- Rebase

dobromir-hristov:
Hey, sorry for the long wait. I am reading this now and I see you are trying to pass a deep object, that is a ref. Isnt that what `reactive` is for? Correct me if I am wrong.
You can try and write a test now, the test suite should be properly working now.

javiertury:
### Scenario

The situation that I have in mind is a component that has to validate an object received from another component. The property could be a bare `ref` instead of a `reactive` object.

There reasons to think the property is not going to be always a `reactive` object, and I think that vuelidate must be able to handle such cases.

### Why such scenario is possible

Functional programming principles require the use of immutable data structures, and `ref` is a better fit for those systems. In vue these principles constitute the base of: the communication between components (props, events) and vuex.

There are alternatives, like the declarative mvvm approach (like mobx-react). In vue this is used to implement the the internal state of components (`data`). This other architecture is better served by `reactive`, but this architecture may not be the best one to solve every possible problem.

All in all, I think it's good to support multiple programming architectures. Then each programmer can choose the most appropriate one to solve their particular problem.

dobromir-hristov:
Yeah, totally. We should be prepared for what the users throw at it. 

Please add a unit test or two and I will see to it as fast as possible. Thanks for your contribution so far :)

javiertury:
The tests are now ready

packages/vuelidate/src/core.js
```diff
@@ -307,7 +307,7 @@ function collectNestedValidationResults (validations, state, key, path, resultsC
   return nestedValidationKeys.reduce((results, nestedKey) => {
     // if we have a key, use the nested state
     // else use top level state
-    const nestedState = key ? state[key] : state
+    const nestedState = key ? computed(() => unwrap(unwrap(state)[key])) : state
```
dobromir-hristov, line 310: Why are we wrapping here in a computed property? Because you would pull the property out of the ref object, making it none reactive?

I wonder if we are wrapping it in a computed any way, do we need to unwrap it?

packages/vuelidate/src/core.js
```diff
@@ -476,22 +477,25 @@ export function setValidations ({
    */
 
   const $model = key ? computed({
-    get: () => unwrap(state[key]),
+    get: () => unwrap(unwrap(state)[key]),
```
dobromir-hristov, line 480: This extra unwrap is because of the computed above, right?

packages/vuelidate/src/core.js
```diff
@@ -476,22 +477,25 @@ export function setValidations ({
...
     }
   }) : null
 
   if (config.$autoDirty) {
-    const watchTarget = isRef(state[key]) ? state[key] : computed(() => unwrap(state)[key])
-    watch(watchTarget, () => {
-      if (!$dirty.value) $touch()
-    })
+    watch(
+      () => unwrap(unwrap(state)[key]),
```
dobromir-hristov, line 495: This is already a computed, because we wrapped it earlier, correct?

dobromir-hristov (commented):
Looks good to me. Left a few small comments and want @shentao to take a final look as well.

packages/vuelidate/src/core.js
javiertury: Yes, the state can be a `ref` because of the `computed` above, or because the user passes a `ref`.

packages/vuelidate/src/core.js
javiertury: Exactly, the computed is needed to preserve reactivity.

`computed` can not track dependencies if they are not unwrapped. There are 2 situations:
  1 - state is a ref (this is a deeply nested property and the parent is `computed`)
  2 - state[key] is a ref (from `useVuelidate(rules, { key })`)

We need to unwrap the refs for every possible case.

packages/vuelidate/src/core.js
javiertury: Yes, if we were using typescript `state` would be of type `T | Ref<T> | Record<keyof T, Ref<T[keyof T]>>` everywhere.

dobromir-hristov (approved):

PR #705, opened 2020-09-27
refactor(core)!: drop mixin, need to use useVuelidate for options api

shentao:
To enable nested validations for Options-API syntax we need to use the same provide/injection logic as in Composition-API. This is pretty hard to accomplish with the Options-API. Given that we can read the validations component property and state of the component from within `setup` we can mix `useVuelidate` with the `validations` component property.

This means the `VuelidateMixin`, and thus the `VuelidatePlugin`, become obsolete and are removed. 

To compare, here’s the current usage example:

```js
export default {
  mixins: [VuelidateMixin], // or have it installed globally as app.use(VuelidatePlugin)
  data: () => ({ x: '' }),
  validations: {
    x: { required }
  }
}
```

And new one:

```js
export default {
  setup () {
    return { vv: useVuelidate() } // vv (or vuelidate) since $ prefixed properties are reserved in Vue 3.0
  },
  data: () => ({ x: '' }),
  validations: {
    x: { required }
  }
}
```

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

**Other information:**

### DIFF
--- a/packages/test-project/src/components/OldApiExample.vue
+++ b/packages/test-project/src/components/OldApiExample.vue
@@ -2,13 +2,13 @@
   <div>
     <label>Number X</label>
     <input
-      v-model="x"
+      v-model.number="x"
       type="number"
     >
     <br>
     <label>Number Y</label>
     <input
-      v-model="y"
+      v-model.number="y"
       type="number"
     >
     <br>
@@ -22,6 +22,21 @@
         {{ error.$message }}
       </p>
     </div>
+
+    <label>Dims</label>
+    <input
+      v-model.number="dims.w"
+      type="number"
+    >
+    <input
+      v-model.number="dims.h"
+      type="number"
+    >
+    <input
+      v-model.number="dims.l"
+      type="number"
+    >
+
     <button @click="vv.$touch()">
       $touch!
     </button>
@@ -43,7 +58,12 @@ export default {
   },
   data () {
     return {
-      y: 3
+      y: 3,
+      dims: {
+        w: 1,
+        h: 1,
+        l: 1
+      }
     }
   },
   computed: {
@@ -52,7 +72,7 @@ export default {
     }
   },
   validations () {
-    console.log(this.y)
+    console.log(this)
     return {
       x: {
         // $autoDirty: true,
@@ -74,6 +94,18 @@ export default {
         isEven: {
           $validator: (v) => v % 2 === 0,
           $message: 'Sum must be an even number'
+        }
+      },
+      dims: {
+        volume: {
+          $validator: (dims) => (dims.h * dims.w * dims.l) > 0,
+          $message: 'Volume must be greater than zero'
+        },
+        w: {
+          minValue: minValue(2)
+        },
+        l: {
+          minValue: minValue(4)
         }
       }
     }
--- a/packages/test-project/src/components/OldApiExample.vue
+++ b/packages/test-project/src/components/OldApiExample.vue
@@ -15,31 +15,31 @@
     <p>X + Y = {{ xPlusY }}</p>
     <div style="background: rgba(219, 53, 53, 0.62); color: #ff9090; padding: 10px 15px">
       <p
-        v-for="(error, index) of $v.$errors"
+        v-for="(error, index) of vv.$errors"
         :key="index"
         style="padding: 0; margin: 5px 0"
       >
         {{ error.$message }}
       </p>
     </div>
-    <button @click="$v.$touch()">
+    <button @click="vv.$touch()">
       $touch!
     </button>
-    <pre style="color: white">{{ $v }}</pre>
+    <pre style="color: white">{{ vv }}</pre>
   </div>
 </template>
 
 <script>
 import { ref } from 'vue'
-import { VuelidateMixin } from '@vuelidate/core'
+import useVuelidate from '@vuelidate/core'
 import { minValue } from '@vuelidate/validators'
 
 export default {
-  mixins: [VuelidateMixin],
   setup (props, context) {
     const x = ref(1)
+    const vv = useVuelidate()
 
-    return { x }
+    return { x, vv }
   },
   data () {
     return {
--- a/packages/vuelidate/src/index.js
+++ b/packages/vuelidate/src/index.js
@@ -1,4 +1,4 @@
-import { provide, inject, ref, computed, getCurrentInstance, onBeforeUnmount } from 'vue-demi'
+import { reactive, toRefs, provide, inject, ref, computed, getCurrentInstance, onBeforeUnmount } from 'vue-demi'
 import { unwrap, isFunction } from './utils'
 import { setValidations } from './core'
 
@@ -14,6 +14,22 @@ const VuelidateRemoveChildResults = Symbol('vuelidate#removeChiildResults')
  * @return {UnwrapRef<*>}
  */
 export function useVuelidate (validations, state, registerAs) {
+  if (!validations) {
+    const instance = getCurrentInstance()
+    if (instance.type.validations) {
+      const rules = instance.type.validations
+
+      state = computed(() => toRefs(reactive(instance.ctx)))
+      validations = computed(() => isFunction(rules)
+        ? rules.call(state.value)
+        : rules
+      )
+
+      const uid = instance.uid || instance._uid
+      registerAs = `_vuelidate_${uid}`
+    }
+  }
+
   // if there is no registration name, add one.
   if (!registerAs) {
     const instance = getCurrentInstance()
@@ -23,6 +39,7 @@ export function useVuelidate (validations, state, registerAs) {
     const uid = instance.uid || instance._uid
     registerAs = `_vuelidate_${uid}`
   }
+
   const resultsCache = new Map()
 
   const childResultsRaw = {}
@@ -80,48 +97,6 @@ export function useVuelidate (validations, state, registerAs) {
       ...childResults.value
     }
   })
-}
-
-/**
- * Vuelidate mixin, used to attach Vuelidate only to specified components
- * Relies on `validations` option to be defined on component instance
- * @type {ComponentOptions}
- */
-
-export const VuelidateMixin = {
-  computed: {},
-  beforeCreate () {
-    const resultsCache = new Map()
-    const options = this.$options
-    if (!options.validations) return
-
-    if (options.computed.$v) return
-
-    const validations = computed(() => isFunction(options.validations)
-      ? options.validations.call(this)
-      : options.validations
-    )
-    let $v
-
-    options.computed.$v = function () {
-      if ($v) {
-        return $v.value
-      } else {
-        $v = computed(() => setValidations({ validations, state: this, resultsCache }))
-        return $v.value
-      }
-    }
-  }
-}
-
-/**
- * Default way to install Vuelidate globally for entire app.
- * @param {Vue} app
- */
-export const VuelidatePlugin = {
-  install (app) {
-    app.mixin(VuelidateMixin)
-  }
 }
 
 export default useVuelidate
--- a/packages/vuelidate/src/index.js
+++ b/packages/vuelidate/src/index.js
@@ -24,9 +24,6 @@ export function useVuelidate (validations, state, registerAs) {
         ? rules.call(state.value)
         : rules
       )
-
-      const uid = instance.uid || instance._uid
-      registerAs = `_vuelidate_${uid}`
     }
   }
 
--- a/packages/test-project/src/App.vue
+++ b/packages/test-project/src/App.vue
@@ -5,6 +5,7 @@
         <li><router-link to="/">Simple Form</router-link></li>
         <li><router-link to="/old-api">Old api</router-link></li>
         <li><router-link to="/nested-validations">Nested Validations</router-link></li>
+        <li><router-link to="/nested-ref">Nested Ref</router-link></li>
       </ul>
     </div>
     <router-view />
--- a/packages/test-project/src/components/ChainOfRefs.vue
+++ b/packages/test-project/src/components/ChainOfRefs.vue
@@ -1,0 +1,57 @@
+<template>
+  <div class="">
+    <input
+      v-model="form.foo.bar"
+      type="text"
+    >
+    <button
+      type="button"
+      name="button"
+      @click="vv.$touch()"
+    >
+      $touch
+    </button>
+    <pre>{{ vv }}</pre>
+  </div>
+</template>
+
+<script>
+import { ref } from 'vue'
+import useVuelidate from '@vuelidate/core'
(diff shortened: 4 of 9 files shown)


packages/test-project/src/components/OldApiExample.vue
```diff
@@ -15,31 +15,31 @@
     <p>X + Y = {{ xPlusY }}</p>
     <div style="background: rgba(219, 53, 53, 0.62); color: #ff9090; padding: 10px 15px">
       <p
-        v-for="(error, index) of $v.$errors"
+        v-for="(error, index) of vv.$errors"
```
dobromir-hristov, line 18: So are we keeping it v or vv? Vv is like veevalidate

packages/vuelidate/src/index.js
```diff
@@ -14,6 +14,22 @@ const VuelidateRemoveChildResults = Symbol('vuelidate#removeChiildResults')
  * @return {UnwrapRef<*>}
  */
 export function useVuelidate (validations, state, registerAs) {
+  if (!validations) {
+    const instance = getCurrentInstance()
+    if (instance.type.validations) {
+      const rules = instance.type.validations
+
+      state = computed(() => toRefs(reactive(instance.ctx)))
```
dobromir-hristov, line 22: Why 'reactive' then 'ref'?

packages/vuelidate/src/index.js
```diff
@@ -14,6 +14,22 @@ const VuelidateRemoveChildResults = Symbol('vuelidate#removeChiildResults')
...
+      validations = computed(() => isFunction(rules)
+        ? rules.call(state.value)
+        : rules
+      )
+
+      const uid = instance.uid || instance._uid
+      registerAs = `_vuelidate_${uid}`
```
dobromir-hristov, line 29: We already do this below, might clean it up a bit.

packages/test-project/src/components/OldApiExample.vue
shentao: Oh dunno. Just needed something to skip the `$` prefix.

packages/vuelidate/src/index.js
shentao: Just using `toRefs(instance.ctx)` works but it throws a warning that it expected a reactive object and not a plain one...

packages/vuelidate/src/index.js
shentao: Done

xanf:
While I definitely understand this from developer perspective, I find an overall approach extremely disturbing. One of the major Vue3 selling points was that "The Composition API is purely additive" (c). While `vuelidate` is not a part of the core, implementing such approach will force people to get into the composition api, making any app which use `vuelidate` to jump onto composition api train even if they do not want to

shentao:
That's a good point! That's also why the setup part is really minimal.
It could really be a one-liner: `setup: () => ({ vv: useVuelidate()})`. 

Sadly we can't have that in a mixin (a design decision in Vue I agree with). I was exploring a mixture of beforeCreate, computed and created to accomplish the same thing without relying on `setup` but sadly that proved only partially working and still causing a lot of problems and seemed super hacky. 

I figured that this approach would be the least intrusive (migration to new Vuelidate) and still allow for easier maintenance in the long run.

packages/vuelidate/src/index.js
dobromir-hristov: Wont this work just fine?
```suggestion
      state = reactive(instance.ctx)
```

shentao: I'm pretty sure I tried that and it did not. :(

desislavsd:
@shentao 

If usage of `setup` is required, won't it be better to remove the validations option as well and provide the validations as an argument to `useVuelidate`? It feels more consistent.

dobromir-hristov:
@desislavsd that is the default behavior

shentao:
@desislavsd 
Yup! The reasons for this change are:
* Bringing the `validations` component-option-API syntax to the same functionality level as when using `useVuelidate` in `setup`. Before this, with the mixin, the validations were not registered/sent to the ancestor component thus not allowing for nested validations support. I’ve tried setting it up using the mixin only, but sadly that proved very hacky and unstable. That’s why I ended up ditching the mixin. With that change, both syntaxes have the same capabilities.
* Keeping the `validations` component-option is important for backward compatibility. If we were to force moving the validation rules into `setup` it would also mean everyone has to move their data to `setup`. And that might require a lot of refactoring. I imagine just having to add `setup: () => ({ vv: useVuelidate() }),` into a component with Vuelidate is not such a big deal, especially if someone hasn’t installed Vuelidate globally as a plugin but relied on adding the mixin where needed. This means the user won’t be forced to use the composition-api for anything else beyond activating vuelidate if they don’t want to.

But if someone wants to go full composition-API, they can skip the `validations` component-option and just use it directly in `setup` like this:

```js
const name = ref('')
const { vv } = useVuelidate({ name: { required } }, { name })
```

desislavsd:
@shentao Thanks for the thorough explanation. I got your point. I had that in mind and still think its possible to move the validations into the setup hook  and access the state ( component data ) without moving the state itself into the setup hook. I know Evan does not encourage it (except for awesome libraries like this one ) but in the setup hook you can access the component instance (currentActiveInstance  / `this` ) . So, as we currently have to pass a function to `validations`  in order to access `data`  properties, the same function may be supported as callback to `useVuelidate()` which then will execute it in the proper context `onCreated()`:

```javascript
data: () => ({ foo: true }),

setup: () =>  ({
  vv: useVuelidate(function(){
    this.foo // true
    return { }
  })
})
```

This way the syntaxis is preserved, only moved to a potentially more proper place :)

aethr:
It seems as though this is a problem with how `instance.ctx` is made reactive and passed to `setValidations`. As `state` gets passed down through `collectNestedValidationResults`, it eventually gets to a point where `state[key]` becomes undefined. Then when this gets used in the generation of the `$model`, the validation loses access to the reactive values.

I was able to fix the issue locally by unwrapping `state` in the first line of `setValidations`:

```
export function setValidations ({
  validations,
  state,
  key,
  parentKey,
  childResults,
  resultsCache
}) {
  state = unwrap(state);
  const path = parentKey ? `${parentKey}.${key}` : key
```

This doesn't seem to affect normal use of the plugin using `setup`, everything that I tested still seemed to work. I was also able to remove the `toRefs` wrapping state on line 22 of `useVuelidate` so it became:

```
      state = computed(() => reactive(instance.ctx));
```

I'm not sure if this is actually a good solution to the issue, but maybe it will give you some more insight into the issue with nested validations.

shentao:
Thank you for the investigation @aethr! Will work on this some more during the weekend and hopefully, we can get this solved soon!

aethr:
Did some testing of this after your changes over the weekend, and I can confirm that the new version is working with some pretty complex validations in my existing project. Nice work!

I did find one difficult bug that took a lot of time to debug. I tried to use the new setup method on an example form that we have in our project for testing validations. This is very similar to your `OldApiExample` component, but I ended up seeing the same errors (undefined keys in `state`) that were occurring before you fixed the state unwrapping issue.

Eventually, I realised it was because I had a `console.log(vv.someProperty)` inside my `setup()` method, like:

```js
  setup (props, context) {
    const x = ref(1)
    const vv = useVuelidate()

    console.log(vv.dims.l); // breaks everything

    return { x, vv }
  },
```

During the `setup` lifecycle `instance.ctx` doesn't actually seem to have the component's `data` expanded/bound yet. If I try to inspect the validation state at this point, it triggers `setValidations` but there's no data inside `state` to validate. While `setValidations` crawls the structure of the validation rules, it simultaneously expands the structure of `state` and attaches the state to the validator using `$model`. But if there isn't a matching branch in `state` it causes an error `Cannot read property 'l' of null` and everything breaks.

If we remove the `console.log` it works fine, I think because the first time validation rules are evaluated is when the template renders, when `data` has been fully attached to the component. I'm guessing if we didn't do lazy evaluation, this problem might be a bigger issue for people using this bridge from the Options API.

As a side note, this would also affect anyone who defines a validation against a branch of state that doesn't exist, but only at 2 levels deep, so for example I broke `SimpleForm.vue` by adding this to validations:

```js
        missingParent: {
          missingChild: { required, minLength: minLength(4) },
        },
```

Perhaps there should be a guard against missing state when setting up `$model`? If `setValidate` is called and `state` or `state[key]` is undefined, I think the validation should be marked as `$invalid` but otherwise business as usual. This seems to be what happens in Vuelidate 1.x.

What do you reckon?

shentao:
Hey! As a small update on the lazy validations – I spent the whole Saturday on it and managed to "unlazy" the validations and introduce a global config that accepts `$lazy` and `$autoDirty` params, where `$lazy` is set to `false` by default. This should restore the old Vuelidate’s behavior.
However, I also hit a major issue related to the Options API – similarly to what you discovered. There are 2 problems that we are facing now:

1. `instance.ctx` doesn’t contain the component’s state the first time it is being called. As you already discovered, this makes it hard to actually watch for changes happening there. This is somewhat "fixed" by having the validation results be a computed property inside `useVuelidate` (the line that calls `setValidations`) except...
2. it shouldn’t be and it wasn’t initially. Although it somewhat solves the 1. issue, it introduces another, much serious one. Whenever the state changes, the `setValidations` is being re-executed which creates new computed properties and watchers. I’m not exactly sure what happens to the computed properties that are no longer used (hopefully they are garbage collected), but the watchers stay. This can lead to unexpected behaviors and memory leaks. It’s more than likely that it also removes all the performance benefits of having all the other Vuelidate properties be their own computed properties. Thus we need to revert back to `const validationResults = setValidations({`. I started working on that [here](https://github.com/vuelidate/vuelidate/blob/unlazy-the-validations/packages/vuelidate/src/index.js#L102) along with restoring the lazy behavior. It works well with `setup` usage but completely breaks the Options-API... Been experimenting with different approaches (hence the WIP commit), but so far no luck.

> Perhaps there should be a guard against missing state when setting up $model? If setValidate is called and state or state[key] is undefined, I think the validation should be marked as $invalid but otherwise business as usual. This seems to be what happens in Vuelidate 1.x.

That’s one of doing it. Alternatively, we just remove `$model` and suggest people use `$autoDirty` that can be passed along specific validation rules like this:
```js
   missingParent: {
          missingChild: { required, minLength: minLength(4), $autoDirty: true },
        },
```

What do you think?
Either way, we would still need a way to protect the user from accessing a deeply nested property that is not yet available.

aethr:
Hey no problem, this is a library my team is using in our product and we're happy to give some time back to help get things moving towards a stable release for Vue 3. Thanks for all the work you've already done!

And, I apologise in advance for the wall of text I'm about to write.

I've been bashing away at this and experimenting with a bunch of different approaches, and I've had a look at your "unlazy" branch as well. Unfortunately I haven't found an elegant solution to any of the problems you're facing. In fact it seems like `useVuelidate` is fundamentally at odds with the Options API for two reasons that you've already discovered:
1. usage of `inject` / `provide` by `useVuelidate` to coordinate parent/child validation passing is *only* possible during `setup`
2. `instance.ctx` is *not* ready during `setup`, meaning we can't build the `$v` object synchronously at this time, if data is defined using Options API

I can see two possible alternatives for keeping Options API support (which I think is important since the Composition API is opt-in and not everyone will want to switch).

**Option 1: Deferred setup via reactivity**

This is the approach you've been working on, but there's some pitfalls. First of all, it seems to add totally unnecessary complexity for anyone who's going all-in on `setup()`. In my previous comment I said:
> this would also affect anyone who defines a validation against a branch of state that doesn't exist, but only at 2 levels deep

and

>  there should be a guard against missing state

The more I think about it, the less convinced I am this is something that should be supported. The Vuelidate implementation will be much simpler (and easier to maintain and less bugs) if validations are required to be defined against a state structure that exists at setup time. As Vue developers we're used to stubbing out the structure of `data` ahead of time, and this feels similar. The upside would be we could run `setValidations` *only once* per component and keep the reactivity really tight around the model and validation params. Keen to know your thoughts on this opinion.

Now, ignoring the above and assuming we *did* want to support deferred setup to solve the Options API problem, we effectively have to watch the *entire* component instance for changes. The simple implementation is to run `setValidations` again each time we see a state change, but wouldn't this be likely to happen really frequently? And we're not just running the validators again, we're re-setting up the entire vuelidate structure each time unless we add some careful plumbing. I can see you've already had to deal with this with when you added `resultsCache`, because otherwise you'd lose the `$dirty` state of every validation each time the state changes.

Ok so what if we tried to be smarter and only run `setValidations` once (or at least more conservatively) when the instance is "ready" or when there's major changes? But then I'm not sure how to determine when that is. Do `data`, `computed` etc all get attached all at once? Is there a nice way to watch/unwatch? My gut tells me probably not, and doing something "clever" here would lead to a lot of accidental complexity, and probably bugs.

Solving the watchers problem is probably possible (maybe using a similar Map[path] strategy you're using with `resultsCache` would allow us to store and call the unwatcher when we are setting up the same path again), but I worry about the added complexity that this entire strategy would bring. I'm starting to feel it's probably not the right approach.

**Option 2: Keep the Options API mixin separate**

I know you attempted this previously and already had a lot of problems that I'm not aware of, apologies if I'm re-treading familiar ground here.

One major problem I am aware of is that the existing Vuelidate 0.x mixin attaches `$v` during `beforeCreate`, which has exactly the same problem we have with `setup`: `data` and `computed` aren't even available yet! I don't know a lot about the implementation there but it appears the solution was to define `$v` using a Proxy, and the validations actually get set up a bit later when the data becomes available. Even if we could make this work for Vuelidate 2.x, I'm not sure it will be compatible with the child validation tracking using provide/inject if it gets set up at the same time.

Another complication is the `provide` / `inject` stuff for child components. Composition API is nice for this since we can add `provide` and `inject` only in the components that call `useVuelidate`. In a mixin in the Options API, I don't think it's easy to conditionally define provide/inject properties only when the `validations` option is present. Maybe it's acceptable to have users include a mixin only for components that do have validations? Seems like an annoyance, maybe it's solvable but I'm not sure.

I'm not sure what other complications you ran into with the mixin, but I do understand that having two parallel implementations is not ideal and creates a maintenance burden. At this stage though, I'm seriously considering whether there might actually be less code to maintain if we go with the mixin, than if we try and support deferred state binding in `useVuelidate`.

Sorry to say, I'm not sure what the best path forward is. I have a fork of the project that I'll try to make public where I'm doing my own experimentation, if you're interested maybe we could find a time to chat in real time and compare notes. Otherwise if you'd prefer to stay focused on the work in this branch, I can try and experiment with the watch/unwatch problem and help find a solution there.

Thanks again for your work on this, and sorry for the massive comment!

dobromir-hristov:
This is already implemented in #790

PR #708, opened 2020-10-01
feat(validators): add "phoneNumber" validator

Roninii:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
> It's submitted to the `dev` branch for v1.x (or to a previous version branch), _not_ the `master` branch

It's submitted to the `next` branch

### DIFF
--- a/packages/validators/src/raw/__tests__/phoneNumber.spec.js
+++ b/packages/validators/src/raw/__tests__/phoneNumber.spec.js
@@ -1,0 +1,49 @@
+import phoneNumber from '../phoneNumber'
+
+describe('phone validator', () => {
+  it('should not validate letters', () => {
+    expect(phoneNumber('abcdef')).toBe(false)
+  })
+
+  it('should validate null', () => {
+    expect(phoneNumber(null)).toBe(true)
+  })
+
+  it('should validate undefined', () => {
+    expect(phoneNumber(undefined)).toBe(true)
+  })
+
+  it('should not validate strings', () => {
+    expect(phoneNumber('hehexd')).toBe(false)
+  })
+
+  it('should not validate space', () => {
+    expect(phoneNumber(' ')).toBe(false)
+  })
+
+  it('should not validate incomplete numbers', () => {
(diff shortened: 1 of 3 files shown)


AllanOricil:
Phone numbers are hard to validate because they vary a lot across multiple countries.
We usually use libs to apply masks, like this https://www.npmjs.com/package/vue-the-mask,  or libs that handle multiple country phones like this https://louismazel.github.io/vue-phone-number-input/

What I would suggest is to have a list of multiple Regex for each country and the Validator would accept a parameter to choose the country or countries. This input would be the official ISO codes.

````
phone: phone('BR')
````
````
phone: phone(['BR', 'PT', 'US'])
```

Or

use a external service to validate the emails without regex, but accepting the ISO code as an input for the validator.

What do you think? @dobromir-hristov @shentao @Roninii

Roninii:
I think that approach makes sense. Maintaining multiple regex checks for this lib _might_ be pushing the scope though. Interested to see what the others have to say about it.

dobromir-hristov:
I think we should leave the phone validator for now. It is too specific to be a simple validator that fits all.

As discussed, there are entire libraries for that, and most people rely on them already for formatting the inputs. Most of them have validators baked in. 

If you want to go into an adventure and try to cover as many as you can, by providing a nice API for users to use, then you can make a small mini package, like `vuelidate-plugin-phone-validators` or something 😄 

We will link it to docs :)

dobromir-hristov:
@shentao wdyt?

shentao:
I’m also in favor of having a separate library for the phone validator that we could recommend in the docs. Given how broad is the topic, I’m afraid we would end up with a constant stream of issues related to certain phone number formats, that would detract us from working on Vuelidate core.

Roninii:
After all the discussion, I think everyone is in agreement. The original estimate of the scope was way smaller. Going to close this.

Issue #707, opened 2020-09-28
Vue 3 + vuelidate@next: $v.$error remains false


jim-cooper:
I'm using Vue 3 and Vuelidate@next and while $v.$invalid works $v.$error and $v.$dirty always remain false even though the individual validators show true (i.e. $v.password.$error returns true).

Sorry, couldn't figure out how to create a jsFiddle with vuelidate@next, but here is my relevant html and js:
```
  <input v-model.trim="$v.email.$model"/>
  <input v-model.trim="$v.password.$model"/>

  data() {
    return {
      email: '',
      password: '',
    };
  },
  validations: {
    email: { required },
    password: { required },
  },
```

Here is my full vue component if needed:
```
<template>
<div>
  <h2 class="header">Sign In</h2>
  $v.$error: {{$v.$error}}<br/>
  $v.$invalid: {{$v.$invalid}}<br/>
  $v.$dirty: {{$v.$dirty}}<br/>
  $v.$anyError: {{$v.$anyError}}<br/>
  $v.$anyInvalid: {{$v.$anyError}}<br/>
  $v.$anyDirty: {{$v.$anyDirty}}<br/>
   $v.password.$error: {{ $v.password.$error }}<br/>
  <form @submit="signIn" autocomplete="off" novalidate>
    <div class="form-group" :class="{ 'error' : $v.email.$error }">
      <label for="email">Email:</label>
      <!-- <em v-if="$v.email.$error && $v.email.email.$invalid">Invalid Email</em> -->
      <em v-if="$v.email.$error && $v.email.required.$invalid">Required</em>
      <input v-model.trim="$v.email.$model"/>
    </div>
    <div class="form-group" :class="{ 'error' : $v.password.$error }">
      <label for="password">Password:</label>
      <em v-if="$v.password.$error">Required</em>
      <input v-model.trim="$v.password.$model"/>
    </div>
    <div class="form-group buttons" >
      <button type="button" @click="cancel()">Cancel</button>
      <button class="save" type="submit" :disabled="$v.$errors.length > 0">
        Sign In
      </button>
    </div>
  </form>
</div>
</template>

<script>
import { required } from '@vuelidate/validators';

export default {
  name: 'SignIn',
  data() {
    return {
      email: '',
      password: '',
    };
  },
  validations: {
    email: { required },
    password: { required },
  },
  methods: {
    signIn(credentials) {
      this.userRepository.signIn(credentials)
        .subscribe(
          null,
[...]

PR #725, opened 2020-10-19
Fix: unlazy-validations with nested objects

MauroB45:
I have been following the discussions about dropping the mixin #705  and making lazy validation optional https://github.com/vuelidate/vuelidate/issues/707  with the new API. Here are some of the things I try to address for this change
- Resolve the issue when building the validation object in `setup` and data is not present in the component
- Refactor the code to centralize the extraction of nestedValidation. Some extra checks were added in order to maintain the correct ref/reactive/computed value throughout the recursive calls
- One of the goals is to only run setValidations a single time and rely on the state reactivity itself to keep things up to date, which should simplify the codebase and make it easier to understand.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
- **NOTE** I realized that when using the Options API to define the validations, the function does not have access to data. This might present some problems if the user tries to access a nested object. Also, for some reason, the computed validation not reacting to changes in the state, and the validation computed object is not updated. 
- Tests can be added to ensure nested validations using OptionsApi does not break the code. I can add some tests if you are happy with these changes

Related to:
- #705 
- https://github.com/vuelidate/vuelidate/issues/707

### DIFF
--- a/packages/test-project/src/App.vue
+++ b/packages/test-project/src/App.vue
@@ -5,6 +5,7 @@
         <li><router-link to="/">Simple Form</router-link></li>
         <li><router-link to="/old-api">Old api</router-link></li>
         <li><router-link to="/nested-validations">Nested Validations</router-link></li>
+        <li><router-link to="/nested-ref">Nested Ref</router-link></li>
       </ul>
     </div>
     <router-view />
--- a/packages/test-project/src/components/ChainOfRefs.vue
+++ b/packages/test-project/src/components/ChainOfRefs.vue
@@ -1,0 +1,57 @@
+<template>
+  <div class="">
+    <input
+      v-model="form.foo.bar"
+      type="text"
+    >
+    <button
+      type="button"
+      name="button"
+      @click="vv.$touch()"
+    >
+      $touch
+    </button>
+    <pre>{{ vv }}</pre>
+  </div>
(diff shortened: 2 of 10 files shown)


aethr:
@MauroB45 and I did a lot of experimenting and digging into the source for the new Vue reactivity system to arrive at the changes that @MauroB45 implemented. As he mentioned in the PR description, a major change here is relying on a `computed` at each level of state (`nestedState`) to bind to the state object for reactivity, instead of wrapping a `computed` around the actual `setValidations` calls themselves.

The `nestedState` at each level can return undefined if it isn't set on its parent yet, but because the parent `nestedState` is a dep, it can update when the parent model updates. Validators can either run and return an `$invalid` state based on a normal evaluation against `undefined`, or hopefully the dev can mark that validation as `$lazy` if they don't want an incorrect `$invalid` value to be calculated before the data is ready.

MauroB45:
Closing this PR as part of https://github.com/vuelidate/vuelidate/pull/740

MauroB45:
Closing this PR as part of https://github.com/vuelidate/vuelidate/pull/740

PR #735, opened 2020-10-23
fix: copy typescript definitions to dist

javiertury:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

When vuelidate is released in npmjs.com some files are excluded from the bundle, and `index.d.ts` is among them. This problem doesn't happen with local installs, that's why I didn't catch it earlier.

This PR makes sure to copy typescript definitions to `dist` so they are preserved. To this end I introduced a new dependency: `rollup-plugin-copy`.


dobromir-hristov:
Jeez.. sorry bout that.

Hm, we should be able to specify these somewhere, it's not necessary to be in the dist.

javiertury:
And, if you don't mind, I would like to make a small improvement to `state` inference. Now the partial match applies to nested properties too and inference from `ExtractStateLeaf` is undependent of `ChildStateLeafs` under `ChildStateLeafs`.

dobromir-hristov:
I did a bit simpler mechanism :) We should be good now.

PR #739, opened 2020-10-26
feat(core): Support returning none boolean data from validators.

dobromir-hristov:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## Description

Extends the support for returning none boolean validators. Now a validators response can return an object and the data it returns will be attached to the `$params` object. This can be used in Error messages and more.

The validator response can have a form like:

```ts
interface ValidatorResponse {
  $invalid: boolean,
  [key: string]: any
}
```


packages/vuelidate/index.d.ts
```diff
@@ -35,7 +35,12 @@ type Component = ReturnType<typeof defineComponent>;
  *
  */
 
-export type ValidatorFn <T = unknown> = (value: T) => boolean;
+export interface ValidatorResponse {
```
dobromir-hristov, line 38: If someone can double check on the types, because I am really not sure :D

dobromir-hristov:
@shentao this is also ready for review.

shentao (approved):

PR #736, opened 2020-10-25
Unlazy the validations

shentao:
**What kind of change does this PR introduce?** (check at least one)

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

**Does this PR introduce a breaking change?** (check one)

- [x] Yes
- [ ] No

If yes, please describe the impact and migration path for existing applications:

The validations are now no longer lazy. We can pass `{ $lazy: true }` to make them lazy though. Introduced `$silentErrors` to list all errors (`$invalid` being true), while `$dirty` is still `false`.

**The PR fulfills these requirements:**

- [ ] It's submitted to the `dev` branch for v1.x (or to a previous version branch), _not_ the `master` branch
- [ ] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix #xxx[,#xxx]`, where "xxx" is the issue number)
- [ ] All tests are passing: https://github.com/vuelidate/vuelidate/blob/master/.github/CONTRIBUTING.md#development-setup
- [x] New/updated tests are included

If adding a **new feature**, the PR's description includes:

- [ ] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)

**Other information:**

PR #740, opened 2020-10-27
Unlazy the validations + Options API improvements

aethr:

- [x] Bugfix
- [x] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

This PR continues the ideas from #705 and #725 but updated with the changes @shentao has made in #736. The main focus is on making the new approaches in Vuelidate 2.x viable for users of the Options API.

The core problem we're trying to solve is that ideally `setValidations` should only need to be called once on each component that defines `validations` via the Options API. Once `validations` has been parsed recursively and the results object (`$invalid`, `$errors`, etc) has been set up, from that point Vue reactivity primitives should be responsible for keeping the results up to date, without having to run `setValidations` again from the root.

The current implementation wraps the initial call of `setValidations` in a `computed`, which may lead to setting up the validations tree multiple times, and necessitates storing things like `$dirty` state separately so that it survives rebuilds of the validation results.

If this approach looks good to you, we may be able to remove some of those parts and simplify the overall implementation.

Issues that we ran into that we believe have been solved in this PR:

**Data is not yet bound to the instance during `setup` lifecycle**

We solved this by creating a `computed` at each level of the state tree (`nestedState`) that returns `null` if that branch of state isn't available yet. Once the state is added in the `onBeforeMount` hook, the changes flow from the root `computed` outwards, and maintain reactivity from that point.

Each validation result is a `computed` with its local version of `nestedState` as a dep, so results are re-computed when `nestedState` changes, without re-building the whole tree.

This has the additional benefit that validations can be set up for data that doesn't exist yet, and as soon as the data is available, reactivity kicks in and the validation will run. In combination with the new "lazy" option we think this could be useful for setting up conditional validations.

**Many reactivity primitives do not seem to work as expected during `setup`**

We use `computed` heavily because it seems to be the only reactivity primitive that works consistently during the `setup` lifecycle. Having read some of the component code in Vue 3 it seems as though reactivity tracking is *paused* during the `setup` method, but `computed` is not affected because it uses its own `effect` implementation.

This may be a mis-reading of what the Vue 3 code is doing, but after extensive testing, `computed` seems to be the most reliable tool for us.

**The `this` object is not yet available when `validations` is defined as a function**

Also due to state not being available from the Options API during the `setup` method, we also have nothing to work with for validations defined like:
```js
validations() {
  someProp: {
    greaterThan: greaterThan(this.otherProp),
  },
},
```

This is tricky to solve, we need to delay calling `validations` until `onBeforeMount` when `instance.proxy` has all our properties set up. However, even doing so, `this.otherProp` doesn't become a dep of the `greaterThan` validator, because the method is called to generate the actual validator function.

The only way to make this reactive is to pass `this.otherProp` wrapped in a computed, like:
```js
validations() {
  someProp: {
    greaterThan: greaterThan(computed(() => this.otherProp)),
  },
},
```
This seems like an uncomfortable syntax for implementors, so instead of binding the component instance directly as `this`, we wrapped it in a computed "factory". This allows implementors to use the more basic syntax in the first example and still get reactivity.

However, this only works when accessing properties directly. If someone wanted to define a validator like `greaterThan(this.otherProp + 2)`, they would still need to wrap it in a `computed` manually to preserve reactivity:
```js
validations() {
  someProp: {
    greaterThan: greaterThan(computed(() => this.otherProp + 2)),
  },
},
```

**Known issues**

Because the validations aren't setup until `onBeforeMounted` for users of the Options API, attempts to access validation results during `setup` will probably cause an error. However, after the experience of writing this code we believe that Options API and `setup` should probably never be mixed by end users, they just aren't that compatible.

**Closing thoughts**

We (@aethr and @MauroB45) are aware that this PR includes sweeping changes, and this is not our project. We're very keen to collaborate to assist you getting this library ready for Vue 3, and we plan on continuing to use the Options API in our current project with Vuelidate when it's ready. If this type of PR is unhelpful or counter-productive we'd love to know how we can help.

I'd also like to note that we haven't included any new tests in this PR because we weren't sure whether the ideas here would be accepted or not. If you're keen to adopt any part of this approach, we'd be happy to include a bunch of tests around the new behaviour before merging.


dobromir-hristov:
First of all, I want to say huge thanks for your effort! Thank you so much for your time.
We both have been either busy or discouraged lately and have not gotten to solving this yet.

I see you have some pretty good ideas there, thanks allot for taking a hack at it! 

I was just thinking of refactoring away the `unlazy`, `globalOptions` and `removeMixin + optionsAPI suppport` work, as they are now pretty much merged into one PR... At this point this will probably be too much work :/ So I guess it will be one stone 3 pigeions...

I will try out the work you have posted here (I was literally just going over your fork just now 😆 

I see we dont have any new tests, which is fine as you stated, we can add those. But did you see some of the old ones break or other strange behavior get introduced, or the ones that were already broken get fixed? 

Things like the `$deoptimize` parameter, that Damian introduced was just a temporary hack, to make sure things work. I see with your approach this should no longer be the case, correct?

dobromir-hristov:
K so I ran the tests, inspected things a bit. I think there is some issue with dynamic validation rules and child component collection. Could you look into those, as they are pretty important?

aethr:
No worries, glad to be helping! We just want to make sure we're doing it in a way that works for you and isn't making things harder.

> I was just thinking of refactoring away the unlazy, globalOptions and removeMixin + optionsAPI suppport work, as they are now pretty much merged into one PR...

Yeah, we started looking at this based on the removeMixin PR and found a few things that didn't work, but we've tried to keep up to date with @shentao 's changes to minimise the work required to merge.

> Things like the $deoptimize parameter, that Damian introduced was just a temporary hack, to make sure things work. I see with your approach this should no longer be the case, correct?

I believe so, yes. We could incorporate that into this PR if it makes sense, or we could do it one step at a time, whatever is easiest.

> K so I ran the tests, inspected things a bit. I think there is some issue with dynamic validation rules and child component collection. Could you look into those, as they are pretty important?

It's evening here now in Australia, so @MauroB45 or myself will have a look at these in the morning and make sure they get fixed up.

MauroB45:
Hi, I had a quick look at the tests failing and found some problems with nested refs. They are fine now, thanks for taking the time to check these changes.

Next, I will now look into the nested validations problems, however, dynamic validations will be hard to get working without the `$deoptimize` config flag to recreate the validation object based on changes in the validation definition. That will bring back all the concerns around possible memory and performance problems. As this is an optional configuration, I think I can bring back the config and make it work for both Option API and composition API.

aethr:
Ok I chatted with @MauroB45 today who did some work fixing a few bugs we introduced that the tests caught, and adding some additional tests around using the Options API. Thanks for pointing us in this direction. :)

Mauro re-introduced the dynamic validations behaviour for users of the Composition API, but left the Options API usage of `setValidations` non-dynamic. I know that this is a feature from Vuelidate 1.x but I guess I'm struggling to see the value of implementing it in this version of Vuelidate where it seems we already have several ways of solving the same problems.

One approach for conditional validations is for the validators themselves to include the condition, similar to the [Optional validator](https://vuelidate.js.org/#sub-optional-validator) example. In combination with the new lazy validations, this gives developers the freedom to create validations that either don't run until the data is ready to evaluate, or evaluate as invalid when the condition isn't met or the data isn't available. In both cases, the validation *result* is available at component creation in case it needs to be used in the template or in a computed property.

Another solution for dynamic validators would be to simply include the validation in the sub-component that gets rendered when the condition is met. This solves the problem of the validator only *appearing* and evaluating when the condition is met, and makes use of existing functionality that you've already included.

I think to implement dynamic validation requires us to be able to run `setValidation` from the root over and over again because we can't make any assumptions about the shape of the validation object without running `validations()` each time the state changes (which on the Options API will be frequently if we're reacting to changes on `instance.proxy`). It raises all the issues we're trying to avoid of possible memory leaks, having to maintain state separately from the validations object, and less focused reactivity.

If we encourage developers to use the new tools to solve these problems, it means a simpler Vuelidate to understand and maintain. I'm interested to know if you think removing dynamic validations might be an option.

dobromir-hristov:
Hey, good job there. 

Now this just my two cents, but introducing such a massive breaking change, like assuming validations are not dynamic for Object Api, is not OK. We cant expect everyone to move from the Object API to using Composition API, I assure you a very minimal part of us will.

Damian's PR also had some commented out tests, which broke and are currently skipped. All of them are somehow related to dynamic validation rules, which worries me.

Using optional validations is an option, but is no where near as convenient, if you want to use pre-existing validators.

I wonder if we could actually watch the validators function for changes... and re-build the validations tree based on that :/

Let me know if this makes sense.

MauroB45:
Good idea. While I was working on this I did some tests using a watcher and it is possible to listen for `instance.proxy` changes. I think we could rebuild the validations tree only on changes in the validation definition. This way we can get dynamic validations and only rebuild when necessary. I will this today

shentao:
@aethr @MauroB45 First of all, I’m truly amazed by the work you both did and even more by your involvement in the project! 😍

We had a call with Dobromir about the issue last weekend and thought about some workarounds for the issue related to dynamic validation rules.

One of the most promising ideas we had was to revert back to using a computed property to collect the validation results, which currently might lead to memory leaks but solves the issue of dynamic validations. 
However, to avoid those leaks we would have to refactor where the validation results are stored. Instead of them living in the function scope of `setValidations` as they are currently and thus having to recalculate every time that function is called, we could resort to storing those similar to the `resultsCache` we have right now. That is, in a separate "module".

I imagine it could look like this:
1. When `setValidations` is first called we create all the current validation results (computed properties, watchers, and so on) and store them in a separate **storage module** created in `useVuelidate`. The results are stored based on data-property+validator pair key or similar that makes them easy to locate and match.
  * The storage module comes with an API that allows us to create and access validation results. 
2. Afterwards we use `getValidations` in a computed would only collect the results of those computations from the storage module. 
  * If there are missing validators repeat step 1, call `setValidations` for the missing rules/properties. Or have them be created automatically, kinda like with missing cache hit.
  * If certain validation results were not accessed – we can remove them from the storage module or keep them if we choose so. It shouldn’t be that hard to track access.
3. This would ensure we only have 1 instance (per component) of all the state, computed and watchers currently repeatedly re-created inside `setValidations`.

As a bonus, it might open up possibilities to specify which properties we want to collect at a time. For example, when someone wants to create 2 forms inside 1 component and those validations should selectively pick nested child component validation results.

Possibly we could also connect the storage to a Vue Devtools plugin once we get to the point where we are ready to build one.

Looking forward to your feedback! Also, if you’d like to discuss it on say Discord or jump on a group call, we’d love to do that! And if you’re interested in a more formalized role in creating and maintaining the library, we could also discuss that.

aethr:
Hey @shentao , some good ideas in here. We were thinking along similar lines a few weeks back to store watchers in a structure like the resultsCache so we can unwatch them when we need to re-create that branch of validations. But as we're moving towards more and more parts of the result structure being cached, I agree it probably makes sense to store the entire result object for each node.

@MauroB45 has also found a way to provide access to instance properties inside the `validations()` method that only causes `validations()` to be re-run when its direct deps change. We were worried that *any* change on the instance would cause it to re-run but we think using a `watch` statement we can avoid it.

My only worry is that I'm not sure we can avoid re-parsing and re-generating the entire tree each time, for example in a scenario like:

```
validations() {
  return {
    weight: {
      maxWeight: (this.condition) ? maxWeightAsyncFromServer() : maxWeight(10),
    }
    height: {
      maxHeight: maxHeight(10),
    },
  };
},
```

If `this.condition` changes after the component has rendered, we'll enter a scenario where `validations()` is re-evaluated, we'll find "weight.maxWeight" already exists in the storage module, but we can't just return it because the actual rule will have changed. We would still need to totally replace everything in it except for `$dirty` and re-run the validator. Hopefully Vue/GC is smart enough to clean up references to the old objects / effects.

At the same time, when `validations()` is run again, the `maxHeight` validator will generate a new function with the param `10` in a closure. Even though the function it returns is functionally identical to the original, the reference will have changed, so again we'll have to replace the entire validator for `maxHeight` even though only the unrelated `this.condition` has changed.

Ultimately, I guess this is the price we pay for having dynamic validations. Your method at least means we can manage the old validations to turn off watchers, etc and hopefully avoid memory leaks. And if users don't choose to use dynamic validations, they still get all the benefits of the tight reactivity gains that @MauroB45 has implemented. So I reckon it's worth trying and see!

Would be happy to jump in a Discord any time we're both awake, I'll reach out to exchange details.

MauroB45:
Hey @dobromir-hristov @shentao, as @aethr mentioned we are thinking along similar lines and it would be interesting to continue the conversations. However, I also think we can keep moving with these changes for now to resolve some problems we are experiencing using the options API. With the latest commits, I added back the dynamic validations behavior for the Old API with some tests. Also, all the broken unit tests have been fixed.

A recap of the changes in the PR,
- The Validation definition function can access the instance and It will be run only when instance properties are bound. This will prevent accessing null or undefined values
- Resolve problems when validating deep objects using the Options API at setup and render
- Provide a strategy to use instance properties in validation rules without the need for recreating the validation tree on every update of instance properties
``` js
export default {
    ...
    data() {
        return {
            x: 5,
            y: 10
        }
    },
    validations(vm) {
        return {
            // Bad. This will cause the validation tree to be recreated every time 'y' is updated
            x: { minValue: minValue(this.y) }

            // Good
            // option 1
            x: { minValue: minValue(vm.y) }
            // option 2 (equivalent to option 1)
            x: { minValue: minValue(computed(() => this.y)) }
            // operation ( it can be a computed and used directly with vm)
            x: { minValue: minValue(computed(() => this.y + 2)) }
        }
    }
}
```
- Dynamic validations for the Options API. It uses a watcher that listen to the validation definition to have better control over what can run `setValidations` multiple times

:facepalm: I am sorry for letting this PR grow to include all of these changes. Also, let me know if you want me to help with the tests that were skipped in the `unlazy-the-validations` branch.

shentao:
Hey @aethr @MauroB45! This has been merged along with other changes in #790! Thanks a lot for your contribution and sorry it took us that long to get that finished. I think a lot of the issues have been resolved and we can finally move forward with other features and improvements. And docs.

PR #750, opened 2020-11-09
docs: custom error messages (fix #748)

### LINKED ISSUE #748: Misleading explanation when changing errors messages in the docs
## The text
From [https://vuelidate-next.netlify.app/guide.html#displaying-error-messages](url) 
<blockquote><p>Error messages come out of the box with the bundled validators in <code>@vuelidate/validators</code> package. You can check how change those them over at the <a href="./custom_validators.html">Custom Validators page</a></p></blockquote>

## The problem

After navigating to <a href="https://vuelidate-next.netlify.app/custom_validators.html">Custom Validators page</a>, I still don't know how to change the error messages. 
IMO, it lacks a proper explanation otherwise it may mislead the reader.

aethr:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: Docs update

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/packages/docs/src/custom_validators.md
+++ b/packages/docs/src/custom_validators.md
@@ -168,7 +168,7 @@ const validations = {
 };
 ```
 
-`$message` can be a basic string, or it can take a function that receives an object with the following properties:
+`$message` can also take a function that is reactive to changes in the validator state and model. The `$messages` function receives an object with the following properties:
 
 | Property   |                            |
 | ---------  | -------------------------- |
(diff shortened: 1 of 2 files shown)


aethr:
@shentao or @dobromir-hristov I just wrote this based on reading the code, so I've made a few assumptions.

Would it also be fine for users to define custom validators with messages using the object syntax? I.e.:

```js
    const isEven = {
      $validator: (val) => val.length % 2 === 0,
      $message: 'Length must be even',
    };
```

If so I can add another section.

dobromir-hristov:
Hey, yeah it's fine, but probably better to use the helper.

The explanation is alright as well. We should mention the message is reactive and if props change, it also changes.

shentao (approved):
Love it! 
One question though, should we perhaps switch the order of arguments for this helper?

aethr:
I suppose the order doesn't bother me too much, I think if you have a lot of basic string messages, having that argument first kind of helps to make it self documenting? If you're using a lot of functions that are defined in-line, it might not be as easy to read.

I'd say the most important thing is that it's consistent with any other `with*` helpers like `withParams`.

PR #745, opened 2020-10-30
fix #743 - VuelidatePlugin ts type to accept App argument

aethr:
**What kind of change does this PR introduce?** (check at least one)

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

**Does this PR introduce a breaking change?** (check one)

- [ ] Yes
- [x] No

If yes, please describe the impact and migration path for existing applications:

**The PR fulfills these requirements:**

- [ ] It's submitted to the `dev` branch for v1.x (or to a previous version branch), _not_ the `master` branch
- [x] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix #xxx[,#xxx]`, where "xxx" is the issue number)
- [x] All tests are passing: https://github.com/vuelidate/vuelidate/blob/master/.github/CONTRIBUTING.md#development-setup
- [ ] New/updated tests are included

If adding a **new feature**, the PR's description includes:

- [ ] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)

**Other information:**

PR #752, opened 2020-11-15
Fix: changed type for VuelidatePlugin to Plugin from vue

mycarrysun:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

---

This was broken in https://github.com/vuelidate/vuelidate/pull/745 and should be fixed now

### DIFF
--- a/packages/vuelidate/index.d.ts
+++ b/packages/vuelidate/index.d.ts
@@ -1,4 +1,4 @@
-import { Ref, defineComponent } from 'vue-demi';
+import { Ref, defineComponent, Plugin } from 'vue-demi';
 type Component = ReturnType<typeof defineComponent>;
 
 /*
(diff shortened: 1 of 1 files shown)


aethr:
Nice pickup.

I changed it to App to match VueRouter `use` type signature, but `Plugin` is the correct type here, my mistake!

gregveres:
Your fix for this introduces a compile error when using Vue 2. Plugin is a new type that is introduced in Vue 3 so it is not available in Vue 2.

mycarrysun:
> Your fix for this introduces a compile error when using Vue 2. Plugin is a new type that is introduced in Vue 3 so it is not available in Vue 2. 

From my understanding, the `next` branch is supposed to only work on vue 3. But I'm not a core contributor, maybe someone else knows more about the strategy for versioning.

Issue #589, opened 2019-12-02
Remove ref helper from Vuelidate 2


dobromir-hristov:
- Start Date: (02-12-19)
- Reference Issues:
- Implementation PR: 

# Summary

Remove the `ref` helper from Vuelidate 2's `@vuelidate/validators` package and related validators as it is mostly inconsistent and incompatible with the new build.

Instead of passing strings to reference values or functions, we directly pass a reference to the required properties.

# Basic example

For validators like `sameAs` or `requiredIf` which use the `ref` helper internally, users will have to change the way they use those validators. 

Instead of passing a string with the name of the reference field, they would need to pass the field itself.

```js
data: ()=>({
  form: {
    password: '',
    confirmPassword: '',
    over18: ''
  }
}),
validations(){
  return {
    form: {
      password: { required },
      confirmPassword: { sameAs(this.form.password) },
      over18: { requiredIf(this.someComputedProperty) }
    }
  }
},
computed: {
  someComputedProperty(){}
}
```

# Motivation and Detailed design

### Confusing syntax

In the current Vuelidate 0.x implementation, the `ref` helper when passed a reference as a **string,** it only retrieves properties in the same object scope as the currently validated property.

Check the example below:

```js
{
  form: {
    personal: {
      first_name: '',
      last_name: ''
    },
    business: {
      email: '',
      address: ''
    }
  }
}
```

If we want to have access to `email` from `first_name` we cant do `requiredIf('form.business.email')`, we can only access properties inside the `personal` property.

To achieve that, we had to pass a function, which then has it's `this` scope bound to the Vue instance or use the passed `parentVM` parameter.

Deep property selectors are also not possible with the `string` notation, you need to use the function method, mentioned above. (There is a PR for this, but it only works with simple objects, not with arrays, maps and so on).

### Incompatible with Composition API

As **Vuelidate 2** is now built with [Composition API](https://vue-composition-api-rfc.netlify.com/#api-introduction), and can be used with it, we need to comply with some of the limitations that it brings with it.

1. There is no Vue instance yet inside the `setup` hook, which prevents us from binding a validator's `this` context to it.
2. We do not have a way to find properties inside the parent object.

# Drawbacks

1. Users will have to use a function for `validations`  opposed to a simple object, to have access to component state.
2. Not backwards compatible. A manual rewrite on custom written validators using `ref` is needed + update the usage for `sameAs,requiredIf,requiredUnless`.
[...]

PR #756, opened 2020-11-22
Allow validations passed to useVuelidate to be a function

mycarrysun:
Somewhat related to https://github.com/vuelidate/vuelidate/issues/589 this PR adds the functionality of being able to pass in a function to `useVuelidate` instead of just a rules object. This allows the validations to be calculated for things that are reactive. I needed this functionality for `sameAs` and added a simple test to check that it works as intended.

I saw the test file was rather large with a lot of empty tests and commented out code, so let me know if I need to put the test in a different spot.

Luckily this wasn't a huge change and we can just call `validationsRaw()` inside `sortValidations` if it is a function. Otherwise the original behavior of `unwrap(validationsRaw)` is performed.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] It's submitted to the `next` branch for v2.x (or to a previous version branch), _not_ the `master` branch

**Other information:**

### DIFF
--- a/packages/vuelidate/test/unit/specs/validation.spec.js
+++ b/packages/vuelidate/test/unit/specs/validation.spec.js
@@ -13,6 +13,7 @@ import {
 import { createSimpleWrapper, shouldBePristineValidationObj, shouldBeInvalidValidationObject } from '../utils'
 import { withAsync } from '@vuelidate/validators/src/common'
 import useVuelidate from '../../../src'
+import { sameAs } from '@vuelidate/validators/src'
 
 describe('useVuelidate', () => {
   it('should have a `v` key defined if used', () => {
@@ -33,6 +34,25 @@ describe('useVuelidate', () => {
 
     expect(vm.v).toHaveProperty('number', expect.any(Object))
     shouldBePristineValidationObj(vm.v.number)
+  })
+
+  it('accepts a function as validations', async () => {
+    const password = ref('a special password')
+    const confirmPassword = ref('')
+
+    const { vm } = createSimpleWrapper(() => ({
+      confirmPassword: {
+        sameAs: sameAs(password, 'Password')
+      }
+    }), { confirmPassword })
+
+    await vm.v.$validate()
+
+    expect(vm.v.$invalid).toBe(true)
+
+    confirmPassword.value = password.value
+
+    expect(vm.v.$invalid).toBe(false)
   })
 
   describe('.$touch', () => {
@@ -329,8 +349,8 @@ describe('useVuelidate', () => {
       const { state, validations } = nestedReactiveObjectValidation()
       const { vm } = createSimpleWrapper(validations, state)
 
-      vm.v.level1.level2.child.$model = 3;
-      expect(vm.v.$errors.find(error => error.$propertyPath === 'level1.level2.child')).toBeTruthy();
+      vm.v.level1.level2.child.$model = 3
+      expect(vm.v.$errors.find(error => error.$propertyPath === 'level1.level2.child')).toBeTruthy()
     })
 
     it('keeps `$params` reactive', () => {
--- a/packages/vuelidate/src/core.js
+++ b/packages/vuelidate/src/core.js
@@ -22,11 +22,13 @@ import { computed, reactive, ref, watch, isRef } from 'vue-demi'
 
 /**
  * Sorts the validators for a state tree branch
- * @param {Object<NormalizedValidator|Function>} validationsRaw
+ * @param {Object<NormalizedValidator|Function>|Function} validationsRaw
  * @return {{ rules: Object<NormalizedValidator>, nestedValidators: Object, config: Object }}
  */
 function sortValidations (validationsRaw = {}) {
-  const validations = unwrap(validationsRaw)
+  const validations = typeof validationsRaw === 'function'
+    ? validationsRaw()
+    : unwrap(validationsRaw)
   const validationKeys = Object.keys(validations)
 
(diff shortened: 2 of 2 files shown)


dobromir-hristov:
I see the change isn't huge, but why did you need to pass a function as validations?

mycarrysun:
The `sameAs` validator and others that had access to the `vm` in v2 can't access it anymore with the composition api so this was the closest thing I could think of to the v2 implementation (same as dynamic validations).

In #589 you mentioned this issue as well.

mycarrysun:
Also, if you look at the test you can see how it works with the `sameAs` validator now but there was no way to actually use the `sameAs` validator with the composition api before this change.

packages/vuelidate/test/unit/validators.fixture.js
```diff
@@ -16,3 +17,4 @@ export const isOdd = (v) => v % 2 === 1
 
 export const asyncIsEven = withAsync(toAsync(isEven, 5))
 export const asyncIsOdd = withAsync(toAsync(isOdd, 5))
+export const sameAs = (equalTo) => (value) => unwrap(value) === unwrap(equalTo)
```
dobromir-hristov, line 20: Why not use the original `sameAs` validator?

packages/vuelidate/test/unit/specs/validation.spec.js
```diff
@@ -35,6 +35,25 @@ describe('useVuelidate', () => {
...
 
+  it('accepts a function as validations', async () => {
+    const password = ref('a special password')
+    const confirmPassword = ref('')
+
+    const { vm } = createSimpleWrapper(() => ({
+      confirmPassword: {
+        sameAs: sameAs(password.value)
+      }
+    }), { confirmPassword })
```
dobromir-hristov, line 46: How is this different than this? The value is a `ref` and is already available, so you can pass it, right?

```suggestion
    const { vm } = createSimpleWrapper({
      confirmPassword: {
        sameAs: sameAs(password)
      }
    }, { confirmPassword })
```

packages/vuelidate/test/unit/validators.fixture.js
mycarrysun: I tried to import it from the relative path but it didn't work because it was out of scope of the tsc I think. What is the proper way to import the actual `sameAs` validator?

packages/vuelidate/test/unit/specs/validation.spec.js
mycarrysun: Yup good catch - I will update that. Forgot that `unwrap` would handle the `.value`.

dobromir-hristov: No, i also removed the function wrapping it. I still dont see the reason for it

packages/vuelidate/test/unit/validators.fixture.js
dobromir-hristov: Just import from `@vuelidate/validators` package, should work.

mycarrysun: Ah makes sense :joy: should've guessed!

mycarrysun: Should be good now

packages/vuelidate/test/unit/specs/validation.spec.js
mycarrysun: Ah I didn't see that when I first looked at it. That was the whole change that this PR made. If you want to separate your validations from the component file there is no way to do that with how the hook currently works. Making a function allows you to pass in whatever you need from the `setup()` function. Here is an example (but let me know if there is a way to accomplish the same thing with the current version and maybe this PR is not necessary, I may just be misunderstanding how hooks work in Vue :sweat_smile: ):

```js
// in my_component.vue
import { UserValidations } from './user-validations';
import useVuelidate from '@vuelidate/core';
export default defineComponent({
  setup(){
    const user = reactive({ email: null, password: null, passwordConfirmation: null });
    const $v = useVuelidate(UserValidations(user), user);
  }
});

// in user-validations.js
import { email, required, sameAs } from '@vuelidate/validators';

export const UserValidations = ({password}) => () => ({
  email: { email },
  password: { required },
  passwordConfirmation: {
    sameAs: sameAs(password, 'Password')
  }
});
```

mycarrysun: And actually typing this out I think it actually might work without return a function :facepalm:

mycarrysun: Maybe not...I just tested it in my project and it is not using the reactive value

dobromir-hristov: Check your validations. You cannot distructure a reactive data. You have to transform it to refs.

mycarrysun: My mistake! Still new at the composition api :sweat_smile: thanks for setting me straight though. I just tried it using `toRefs` and it worked! :tada:

mycarrysun:
Closing because this was not necessary

packages/vuelidate/test/unit/specs/validation.spec.js
shentao: Yay! Glad to see it worked! Totally understandable the composition-API has it quirks.

Issue #753, opened 2020-11-18
Capital letters in domains are invalid


WhiskeeDev:
https://jsfiddle.net/rfdsvh0m/

Not much extra information I can give for this one, title and jsfiddle cover technical.

`email@domain.com`: Valid
`email@Domain.com`: Invalid

aethr:
@WhiskeeDev this was brought up a couple of weeks ago in https://github.com/vuelidate/vuelidate/issues/747#issuecomment-720807766

Given that domains are case-insensitive, this seems like a good safety feature, but others may not agree. What's your opinion?

WhiskeeDev:
In terms of preventing duplicate email address, this does seem like a good feature.
However the alternative method of converting a user's input before validation can cause alarm, especially to those less aware of how email addresses how.
From a UX perspective, if I was a user who thought domains _were_ case-sensitive, I could become confused/annoyed when an input field keeps changing my email to lower-case.
And with the way vuelidate works at the moment, it's not possible to validate the email without either forcing a change in the value or cloning the original value (to avoid distracting the user) and then changing that, to be validated.

bstapleton:
Agree that case sensitivity **may** be a concern regarding potential for duplicates, but having it as a toggleable state (perhaps with a default of CI) for developers to apply would allow the best balance between UX and back-end requirements for data formatting.

aethr:
@WhiskeeDev I agree that changing the user's input on the fly would not be a good UX, I would probably create a computed property with a sanitized email to use for validation and storage. Perhaps mentioning this in the docs would help developers who run into this design decision?

Maybe it's not the library's job to enforce how people manage emails in their systems, but I think one of the reasons people reach for a library like Vuelidate is to make doing "the right thing" easier. This feels like one of those cases.

aethr:
> Agree that case sensitivity **may** be a concern regarding potential for duplicates, but having it as a toggleable state (perhaps with a default of CI) for developers to apply would allow the best balance between UX and back-end requirements for data formatting.

For the most part, I think Vuelidate has gone with the approach of making custom validators very easy to write instead of trying to accommodate everyone. You could probably make a validator that did this by wrapping a function around the existing email validator and transforming the host to lowercase before passing it through, leaving the source data unchanged.

aethr:
Honestly, I'm not sure about this one. If you want to normalize how you store email domains, it has to be done on the server at the end of the day, it's probably not the job of your validation library.

[...]

PR #762, opened 2020-12-01
Fix email validator bug (uppercase domain)

shin10:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

The URL validator has been updated 3 years ago to accept mixed case domains:
https://github.com/vuelidate/vuelidate/issues/220

Somehow this has not been changed in the email validator, yet. This PR adds the ignore case modifier to the regex.


dobromir-hristov:
before I merge this, do we have a concrete source, that emails should be case-insensitive? I thought that the part prior `@` is case-sensitive, and the domain after that isnt?

shin10:
I found this:

> The local-part of a mailbox MUST BE treated as case sensitive.
   Therefore, SMTP implementations MUST take care to preserve the case
   of mailbox local-parts.  In particular, for some hosts, the user
   "smith" is different from the user "Smith".  However, exploiting the
   case sensitivity of mailbox local-parts impedes interoperability and
   is discouraged.  Mailbox domains follow normal DNS rules and are
   hence not case sensitive.
https://tools.ietf.org/html/rfc5321#section-2.4

In other words, you are right, especially for mail delivery.

For the validation of an input field however it means both (local-part and domain) can include lower and/or upper case letters.

aethr:
@dobromir-hristov following the discussion in https://github.com/vuelidate/vuelidate/issues/753 I think it makes sense **not** to enforce lowercase on email domains in Vuelidate.

Because the domain is case-insensitive, `EXAMPLE.COM` and `Example.com` technically are both valid ways to express `example.com`. While I strongly feel that the domain part should be made lowercase *during validation against duplication* (and probably storage), that is probably a job best handled server-side and not in the front-end validation rules.

shin10:
@dobromir-hristov 
Is there anything else, I can/should do?

dobromir-hristov (approved):

bryce-mcmath:
@shin10 Do you know what release this change is/will be in?

eduarguz:
@dobromir-hristov sorry to bother... May we have a release with this fix? I think this has not been tagged
Thanks!

malinushj:
For those looking for an intermediary fix, here's my solution using a helper function:

```js
// force the domain of the email to be lowercase and send it to the email validator
export const emailFormatter = (value) => {
  if (!value) return value;
  let parts = value.split('@');
  parts[1] = parts[1].toLowerCase();
  return parts.join('@');
}
```

This is how it would be used:
```js
import { email, required } from 'vuelidate/lib/validators';
import { emailFormatter } from ...

// ...
validations: {
    form: {
      email: {
        required,
        email: (val) => email(emailFormatter(val)),
      },
    },
  },
// ...
```

Issue #773, opened 2020-12-23
Invalid TypeScript declaration for sameAs


khorsky:
I'm using TypeScript, Vue 3, vuelidate/core 2.0.0-alpha.8, vuelidate/validators 2.0.0-alpha.5.

Code and rules work as expected, but I'm encountering TypeScript error "Expected 2 arguments, but got 1" for sameAs validator. Rule definition: 

```
// ...

controlPassword: {
  sameAs: sameAs(newPassword),
},

// ...
```

When trying to find out where do I have a bug, I found file `node_modules/@vuelidate/validators/index.d.ts`, that exports this:

```
export const sameAs: <E = unknown>(
  equalTo: E,
  otherName: string
) => ValidationRuleWithParams<{ equalTo: E, otherName: string }>; 
```

I believe that's the source of error, but I'm not sure if it's not intended for some reason...

dobromir-hristov:
Please contribute a fix, TS was not first priority at the time of writing the alpha version.

PR #776, opened 2020-12-26
fix: declare second sameAs validator parameter as optional (#773)

khorsky:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: TypeScript declaration fix

- [ ] Yes
- [x] No

**Other information:**
Forgotten question mark, I guess...

### DIFF
--- a/packages/validators/index.d.ts
+++ b/packages/validators/index.d.ts
@@ -42,6 +42,6 @@ export const requiredIf: (prop: boolean | string | (() => boolean | Promise<bool
 export const requiredUnless: (prop: boolean | string | (() => boolean | Promise<boolean>)) => ValidationRuleWithoutParams;
 export const sameAs: <E = unknown>(
   equalTo: E,
-  otherName: string
+  otherName?: string
 ) => ValidationRuleWithParams<{ equalTo: E, otherName: string }>;
 export const url: ValidationRuleWithoutParams;


dobromir-hristov (commented):
Technically true, though I dont think anyone will want `The value must be equal to the other value`


I think we should force that second parameter

shentao:
I assume this is for the built-in validator already with message, right? So if someone wants to use the `sameAs` validator but without the requirement on the name of the 2nd param, they can use the raw validator directly or build a validator on top of it that has a different message format. 

On the other hand, if someone would like to modify the existing built-in validator with `withMessage` then that 2nd param becomes potentially unnecessary.

khorsky:
Currently I set rules for `sameAs` this way:
```
const rules = {
  newPassword: {
    required,
  },
  controlPassword: {
    sameAs: sameAs(newPassword),
  },
}
// ...
```

TS compiler is complaining about missing second parameter for `sameAs.` I'm sorry if I haven't understand your thoughts, but I'd believe that second parameter should be optional.

shentao:
@dobromir-hristov I guess if someone doesn’t care about the error message then I think we should make it optional. No harm in that.

dobromir-hristov:
Though the same.

PR #782, opened 2021-01-06
ValidationRule should accept boolean to cater $autoDirty: true

kayue:
ValidationRule should accept boolean to cater `$autoDirty: true`

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### DIFF
--- a/packages/vuelidate/index.d.ts
+++ b/packages/vuelidate/index.d.ts
@@ -48,7 +48,7 @@ export interface ValidationRuleWithParams<P extends object = object, T = unknown
   $params: P
 }
 
-export type ValidationRule <T = unknown> = ValidationRuleWithParams<any, T> | ValidationRuleWithoutParams<T> | ValidatorFn<T>;
+export type ValidationRule <T = unknown> = ValidationRuleWithParams<any, T> | ValidationRuleWithoutParams<T> | ValidatorFn<T> | boolean;
 
 type ValidationRuleCollection <T = unknown> = Record<string, ValidationRule<T>>;
 


dobromir-hristov:
You sure about this? A validation rule has to be a function or an object. I dont see how it can be a boolean.

shentao:
I’m also not sure how this should work. @kayue could you elaborate a bit more about this use case?

kayue:
`$autoDirty: true` is one of the ValidationRule. 

Please see https://vuelidate-next.netlify.app/guide.html#the-dirty-state

dobromir-hristov:
It's a property, but the entire validation cannot be a boolean. Maybe it's me and I don't get how we should refactor the type better.

dobromir-hristov:
I am going to close this. A validator can two things, sorry of my TS lingo is wrong here... But it cannot be just a boolean

```ts
const Validator = Function

interface Validator {
  $validator: Function,
  $autoDirty?: Boolean,
  $lazy?: Boolean,
  $params?: Object
}
```

kayue:
@dobromir-hristov Thanks for following up. I thought I was modifying `ValidationRule` (not validator), which in your example it can be "Function, Boolean and Object".

I will cycle back when I can reproduce this, I am in another project now.

Thanks.

kayue:
@dobromir-hristov this is the error that I ran into
Screenshot 2021-03-26 at 12 44 37 AM

dobromir-hristov:
We should tell it that a validator can accept configuration, not a raw boolean.

If you can do that, would be awesome

kayue:
@dobromir-hristov Sorry not an expert in defining type, I don't understand what you meant.

dobromir-hristov:
Then you will have to wait for one of us to fix it. I may try this weekend.

For now, You can just pass the configs as the third parameter to useVuelidate, they will be passed to ass validators.

PR #796, opened 2021-02-05
fix: autoDirty on root, closes #792

### LINKED ISSUE #792: $autoDirty does not work when passing as a globalConfig
Passing `$autoDirty` with a plain object does not seem to work properly, making all nested form properties auto dirty.

```js
{
const v = useVuelidate(
   { name: { required } }, 
   { name: ref('foo') }, 
   { $autoDirty: true }
 )
}
```

dobromir-hristov:
## Summary

Fixes the issues with `$autoDirty` when used as a root level config.


shentao (approved):
LGTM!

PR #808, opened 2021-02-19
refactor: rename $invalid response property to $valid

dobromir-hristov:
## Summary

This PR renames the `$invalid` property on rich validation responses to `$valid` to better reflect it's value.

BREAKING CHANGE: It used to be `$invalid`, but that did not make any sense, as the return value was identical to boolean validators.

### DIFF
--- a/packages/docs/package.json
+++ b/packages/docs/package.json
@@ -16,6 +16,6 @@
   },
   "devDependencies": {
     "tree-kill": "^1.2.2",
-    "vitepress": "^0.11.5"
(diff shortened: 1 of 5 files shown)


shentao (approved):

shentao:
Good catch! Lets merge this

masihtak:
`$isValid` would be much better in my opinion

dobromir-hristov:
We don't use a pattern of isSomething in the library, that's why I went with that.

PR #810, opened 2021-03-04
feat/add current vm to validator context, fixes #775

### LINKED ISSUE #775: parentVm is undefined
"@vuelidate/core": "^2.0.0-alpha.8",
"@vuelidate/validators": "^2.0.0-alpha.5",

withParams(params: Params, rule: CustomRule | ValidationRule)

CustomRule = (value: any, parentVm?: any)

CustomRule doesn't receive the second parameter parentVm

dobromir-hristov:
## Summary

This PR fixes the issue in #775  where we did not pass the parentVM to the validator. I know that we can make this with a validator parameter, but it may be annoying to pass those all the time, and this is a breaking change from Vuelidate 0.x

The fix is padding the publicInstance so we should not be exposing anything out of the ordinary to the user.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**


dobromir-hristov:
@shentao tested, no validators are getting re-created on instance change. We only watch what we need. The only computed we have on a big piece of data is the final unwrap, but that is also not getting called.

shentao (approved):

PR #823, opened 2021-03-14
fix/handle async functions in validators

dobromir-hristov:
## Summary

Fixes an issue with some validators not properly handling async validators.

This PR also refactors the `Custom Validators` page, to have a more readable format, with examples for the more complex validators.


dobromir-hristov:
@shentao I am merging this, as it fixes a potentially nasty issue with Validators, and it improves docs quite a bit.

PR #825, opened 2021-03-14
feat: add higher-order component helper for vuelidate

shentao:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

A helper higher-order component that should make it possible to put validation rules inside the template easily instead of defining those in the wrapper component. Example usage in test-project:

```js
// _TextInput is an input component
const TextInput = withVuelidate(_TextInput)

// in template
<TextInput v-model="value" :rules="{ required }" />
```

**Sad part:** Works but there is an infinitely recursive thing going on which makes everything super slow.

**Other information:**

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     ]
   },
   "dependencies": {
-    "vue-demi": "^0.6.0"
+    "vue-demi": "^0.7.3"
   },
   "optionalDependencies": {
     "@vue/composition-api": "^1.0.0-rc.1"
--- a/packages/components/.browserslistrc
+++ b/packages/components/.browserslistrc
@@ -1,0 +1,6 @@
+# Browsers that we support
+
+last 2 versions
+> 1%
+maintained node versions
+not dead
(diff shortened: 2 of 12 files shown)


dobromir-hristov:
@shentao I looked into this and if the rules are defined as `computed` or `data` properties and passed to the `WithVuelidate` wrapped component, there is no recursion issue. But applied on the component itself, it just does not work

shentao:
Uh... Any idea how to solve that?

dobromir-hristov:
I think it's just a problem with applying a watcher to a reactive prop, defined in the template.
I did something similar today, but with array.reverse in the template. Threw the same error, fixed it by creating a computed property.

shentao:
Could this be a vue 2.x limitation?

dobromir-hristov:
I think it's in both versions.

PR #837, opened 2021-04-04
feat(core): add external validations support

### LINKED ISSUE #824: Displaying external validation errors
**Is your feature request related to a problem? Please describe.**
Vuelidate does a great job with client-side validation, but sometimes we need to implement server-side validation as well. I know it is possible to add custom validations, but what if the only method to find out about errors is to submit a form?

Here's a simple example:
https://codesandbox.io/s/eager-clarke-nmz2q

I would like to be able to nicely display the errors that occurred after the form was submitted. Is it possible? Unfortunately, I was unable to find a solution.

**Describe the solution you'd like**
It would be great to be able to inject temporary error messages that go away as soon as the user interacts with the form.

**Describe alternatives you've considered**
Now we have to use various crutch solutions, for example:
https://blog.philiphendry.me.uk/2020/01/09/integrating-server-side-errors-with-vuelidate
https://stackoverflow.com/questions/55497881/vuelidate-server-side-validation

**Additional context**
https://github.com/vuelidate/vuelidate/issues/277

dobromir-hristov:
## Summary

Adds external validator support to Vuelidate. For each state property, we can have a validation error from the server. It can be a single string or even an array of errors. closes #824 

```
const state = {
  child: 'a',
  nested: {
    nestedChild: 'b'
  }
}

const externalErrors = {
  child: 'Child is not Valid',
  nested: {
    nestedChild: ['Foo', 'bar']
  }
}
```

This adds a `$externalResults` key to the Vuelidate validation object as well as triggers both `$invalid` and `$error` if said property external results are not empty.

## Usage in Composition API

```js
const state = {
  number: 0
}
const $externalResults = ref({});
const v = useVuelidate(rules, state, { $externalResults })

async function addExternalErrors(){
 $externalResults.value = { number: 'Must be above 0' }
}
```

## Usage with Options API

If you need to have external results, just create a `vuelidateExternalResults` object, preferably with the same structure as your state, in the instance data.

```js
export default {
  data: () => ({
    name: 'foo',
    vuelidateExternalResults: { name: 'Must Not be Foo' }
  })
}
```

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/packages/docs/src/advanced_usage.md
+++ b/packages/docs/src/advanced_usage.md
@@ -224,10 +224,10 @@ export default {
 }
 ```
 
-:::tip You can pass validation configs as a single parameter to `useVuelidate`
-
+:::tip
+**Note:** You can pass validation configs as a single parameter to `useVuelidate`
 - [Passing a single parameter to useVuelidate](#passing-a-single-parameter-to-usevuelidate)
-  :::
+:::
 
 ## Returning extra data from validators
 
@@ -331,19 +331,19 @@ export default {
 
 ## Providing external validations, server side validation
 
-To provide validation messages from an external source, like from a server side validation response, you can use the `externalResults` feature. Each
-property in the validated state can have one or array of messages. This works with both Composition API and Options API.
+To provide validation messages from an external source, like from a server side validation response, you can use the `$externalResults` functionality.
+Each property in the validated state can have a corresponding string or array of strings as response message. This works with both Composition API and
+Options API.
 
 ### External results with Composition API
 
-When using the Composition API, you can pass a `reactive` object, or a `ref` object, to the `$externalResults` global config.
+When using the Composition API, you can pass a `reactive` or `ref` object, to the `$externalResults` global config.
 
 ```js
 // inside setup
 const state = reactive({ foo: '' });
-const $externalResults = reactive({})
-// you can also use a ref
-// const $externalResults = ref({})
+const $externalResults = ref({}) // works with reactive({}) too.
+
 const rules = { foo: { someValidation } }
 const v = useVuelidate(rules, state, { $externalResults })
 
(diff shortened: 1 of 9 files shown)


rjd22:
This functionality seems extremely useful. @dobromir-hristov can I help with this?

dobromir-hristov:
I think it was mostly done, just need to double check with latest changes to next branch, and if it works with object API.

dobromir-hristov:
I think we could have probably went without `vuelidateExternalResults` on the OptionsAPI and instead tell people to pass a ref to the `useVuelidate` config instead, but thats mixing the two concepts so I would rather we dont do that.

rjd22:
@dobromir-hristov I also think that having a clear input and output is way more sensible. injecting refs that will be changed feels really familiar to a double dispatch, that might be confusing to most developers.

entermix:
@dobromir-hristov Thanks!

c0depwn:
With what npm package version will this be available? With the current latest version of `^2.0.0-alpha.19` this functionality does not seem to be available. I have just noticed this since the `GlobalConfig` type is missing the `$externalResults`. Is there a workaround or will I just have to wait for the next release?

dobromir-hristov:
I think I have not released it yet. Should be out soon.

c0depwn:
@dobromir-hristov Alright thanks  I'm really looking forward to this feature, great stuff! 🙂 I was a bit disappointed since the docs already have an example of this.

dobromir-hristov:
It should be out now, please give it a try.

c0depwn:
@dobromir-hristov Thank you for the quick response/release! 

Unfortunately I have noticed some breaking issues with the implementation:
1. Using  `const $externalResults = ref({});` as described in the docs and passing it like this  `const v$ = useVuelidate(rules, state, { $lazy: true, $externalResults });` will result in the following type error: `Type 'Ref<{}>' is not assignable to type 'ServerErrors'`. I cannot type`ref<ServerErrors>({})` since `ServerErrors` is not exported.
2. Additionally, `GlobalConfig` now requires the argument `$externalResults` which breaks all existing implementations not using this features, I'm referring to the following:
```
interface GlobalConfig {
  $registerAs?: string
  $scope?: string | number | symbol | boolean
  $stopPropagation?: boolean
  $autoDirty?: boolean
  $lazy?: boolean,
  $externalResults: ServerErrors <-- this should probably be optional ($externalResults?)
}
```

I'm currently using typescript and vue 3. If you want I can create a separate issue. But for now until these problems are addressed I will have to downgrade and use a workaround. I hope this is helpful.

dobromir-hristov:
Please create a PR to fix the TS issues. I cannot do that atm.

PR #846, opened 2021-04-18
feat(core): accept a computed property directly as a validator

dobromir-hristov:
## Summary

Allow using a `computed` as a validator. This is useful when the validator depends on external reactive properties, not just the model.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/packages/vuelidate/src/core.js
+++ b/packages/vuelidate/src/core.js
@@ -1,5 +1,5 @@
 import { isFunction, unwrap, unwrapObj } from './utils'
-import { computed, isRef, reactive, ref, watch, nextTick } from 'vue-demi'
+import { computed, isRef, reactive, ref, watch, nextTick, toRaw } from 'vue-demi'
 
 let ROOT_PATH = '__root'
 
@@ -10,6 +10,7 @@ let ROOT_PATH = '__root'
 /**
  * @typedef NormalizedValidator
  * @property {Validator} $validator
+ * @property {ComputedRef<Promise<ValidatorResponse> | ValidatorResponse>} $computedValidator
  * @property {String | Ref<String> | function(*): string} [$message]
  * @property {Object | Ref<Object>} [$params]
  * @property {Object | Ref<Object>} [$async]
@@ -41,7 +42,7 @@ function sortValidations (validationsRaw = {}) {
   const config = {}
 
   validationKeys.forEach(key => {
-    const v = validations[key]
+    const v = toRaw(validations)[key]
 
     switch (true) {
       // If it is already normalized, use it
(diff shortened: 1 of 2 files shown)


dobromir-hristov:
This works OK for Vue3 but has issues with Vue 2. `toRaw` does not return the original object of `reactive`, so I cannot get the original computed property for a validator. This also causes potential issues when a validation branch is a reactive property:

```
{
  foo: computedValidator(() => someFoo === 'foo'),
  bar: computed(() => conditional ? {} : { validators })
}
```

I think there are a few ways we can go:

1. We return back sync validators via `computed`, as a default way to instantiate validators. This will be a breaking change. Not sure if there are other drawbacks, but the validator will run just when needed.
2. We use the validator as the target for the `watch` function in `createValidatorResult`. This works pretty well, if the validator does not rely on the `value` at all, but users cannot access the current value, which I think is a drawback.
3. As an alternative to (2), we allow the users to specify a `$watchTarget`. This way we can keep using the watcher, and dont worry about async validators, yet still give users flexibility to provide any validator. I am worried this is probably too internal, and we have to merge their target with our `[model, $dirty]` target 😐 Something like this - `validator: { $validator: isFoo, $watchTarget: [foo, bar, baz] }`

shentao:
Hey! Thanks for the detailed explanation. I really have no idea which of those would be the best. I think the best one would likely be 2) since we might not want to revert to 1). What did you mean by not being able to access the `value`?

dobromir-hristov:
Going back to 1), would fix the majority of sync validator issues, leaving async to require special handling, like before.

With 2, we just use the validator as a watcher source, or an alternative to that is 3), allow them to specify a watcher source... albeit that sounds too much :/

```js
  if (rule.$indirectValidator) { // or however we call this
    $unwatch = watch(rule.$validator, (result) => { // because we use the validator as a watch target, we never pass `$model` to it, thats what I meant
      $response.value = result
      $invalid.value = normalizeValidatorResponse(result) // and the result of each call as a result.
    }, { immediate: true })
  } else {
    // this is what we have now
    ({ $invalid, $unwatch } = createAsyncResult(
      rule.$validator,
      model,
      $pending,
      $dirty,
      config,
      $response,
      instance
    ))
  }
```

shentao:
I have no opinion here. Both 1) and 2) seem like OK solutions to me.
I wouldn’t worry that much about async validators and possibly requiring once again the `$async` flag, as those are still rather uncommon so it shouldn’t lead to too many issues.
Still unsure what is the con of 2) and the `value` you mentioned. @dobromir-hristov

shentao:
@dobromir-hristov Then I believe we might want to introduce the mentioned `$watchTargets` for async validators and revert to computed properties for sync ones. Wdyt?

dobromir-hristov:
Might not be a bad idea... I will look into this.

dobromir-hristov:
Preceded by #860

PR #850, opened 2021-04-27
Fix #302: Remove web build conditional

### LINKED ISSUE #302: v0.7+: $params values are null
As you can see in the [default fiddle](https://jsfiddle.net/Frizi/b5v4faqf/) the `$params` values are `null`:
```js
{
  "text": {
    "required": false,
    "minLength": true,
    "$model": "",
    "$invalid": true,
    "$dirty": false,
    "$anyDirty": false,
    "$error": false,
    "$anyError": false,
    "$pending": false,
    "$params": {
      "required": null,
      "minLength": null
    }
  }
```
So using `$v.text.$params.minLength.min` like the docs say gives a type error.

This bug appeared in v0.7.0, it was working in v0.6.2.

kelunik:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

I can't find any version branches, is this template up to date?

This fixes #302 while building with webpack and having `BUILD=web` set as environment variable. The global `vuelidate` object isn't available when building with webpack.


kelunik:
@dobromir-hristov I'm importing from `import { maxLength, maxValue, minLength, minValue, required } from 'vuelidate/lib/validators';`, as documented, but it doesn't work.

`vuelidate/lib/withParams.js` contains:

```
"use strict";

Object.defineProperty(exports, "__esModule", {
  value: true
});
exports.default = void 0;
var withParams = process.env.BUILD === 'web' ? require('./withParamsBrowser').withParams : require('./params').withParams;
var _default = withParams;
exports.default = _default;
```

`process.env.BUILD === 'web'` is available and defined, but chooses the no-op implementation.

kelunik:
As this hasn't been accepted, I've now locally patched it and [documented it on Stack Overflow](https://stackoverflow.com/a/69978156/2373138).

Issue #657, opened 2020-04-29
How can I determine if a value is currently meeting requiredIf standards?


afineedge:
https://jsfiddle.net/gr4vodem/3/

The `required` attribute reflects the success/error state of a field, returning `true` if the field is filled in and `false` if it is not. When a field has `requiredIf` resolving as `true`, that behavior is maintained. However, if requiredIf resolves as false, both an empty field and a field with data have identical attributes, with `required` resolving as true even though the field is not required. In addition, whether `requiredIf` resolves as true or false, the attributes for both fields.

This leaves no obvious way to determine whether a field's `requiredIf` attribute resolves as `true` or `false`. I have a complex form with many fields that are displayed or made required conditionally, so indicators for required fields must display only when needed. Is there a way to determine whether a field currently meets `requiredIf` standards?

afineedge:
I figured out how to get the information I needed through `this.$v.field.$params.required.prop(this)` in the top-level component, and using `this.$v.field.$params.required.call.prop(this.$parent)` in any child component. However, this feels like overkill. When one checks `this.$v.field` on a field with a `requiredIf` in its validation, it would be helpful to see an `isRequired` or equivalent attribute when the `requiredIf` requirements are met.

PR #856, opened 2021-05-13
#657: add isRequired param to required* validators

littlejon85:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:** #657

Often there is a need to determine whether a validated field is required. This PR adds boolean *isRequired* to *$params* of **required**, **requiredIf** and **requiredUnless** validators to indicate wheter the field is actualy required.

This is useful for example to display a required star in the label of the field without the need to use custom validators.

Please decide whether this is appropriate solution to the problem. Also I'd like to check if the name *isRequired* is adequate. Other validators pass $params with the same name as input params eg. **maxValue**.

### DIFF
--- a/packages/validators/src/withMessages/required.js
+++ b/packages/validators/src/withMessages/required.js
@@ -6,5 +6,6 @@ import required from '../raw/required'
  */
 export default {
   $validator: required,
-  $message: 'Value is required'
+  $message: 'Value is required',
+  $params: { isRequired: true }
 }
--- a/packages/validators/src/withMessages/requiredIf.js
+++ b/packages/validators/src/withMessages/requiredIf.js
@@ -9,6 +9,6 @@ export default function (prop) {
   return {
     $validator: requiredIf(prop),
     $message: 'The value is required',
-    $params: { isRequired: prop }
+    $params: { isRequired: !!prop }
   }
 }
--- a/packages/validators/src/withMessages/requiredIf.js
+++ b/packages/validators/src/withMessages/requiredIf.js
@@ -8,6 +8,7 @@ import requiredIf from '../raw/requiredIf'
 export default function (prop) {
   return {
     $validator: requiredIf(prop),
-    $message: 'The value is required'
+    $message: 'The value is required',
+    $params: { isRequired: prop }
   }
 }
--- a/packages/validators/src/withMessages/requiredUnless.js
+++ b/packages/validators/src/withMessages/requiredUnless.js
@@ -7,5 +7,6 @@ import requiredUnless from '../raw/requiredUnless'
  */
 export default prop => ({
   $validator: requiredUnless(prop),
-  $message: 'The value is required'
+  $message: 'The value is required',
+  $params: { isRequired: !prop }
 })


dobromir-hristov:
I am sorry, but I don't see a use case for this. Can you provide one?

littlejon85:
> I am sorry, but I don't see a use case for this. Can you provide one?

Of course. This would be useful for example in libraries to determine whether a conditionally required input fields are actually required. Before vuelidate v2 this was somewhat possible by checking the $params in the root of the $v validation object:

image

However this is not possible in v2. I believe there is need to let the user know whether the requiredIf and requiredUnless validators are actually in effect or not. This would also make these validators conform to other validators which expose their params in $params eg. maxValue.

At this moment the user has no way to distinguish between active and inactive requiredIf validator. This is the solution I propose:

Value is required:
image

Value is not required:
image

I stumbled upon this need when rewriting our components library to Vue 3. Hope I made the use case clear. If you think there is a better way around this please let me know. Thank you.

littlejon85:
Here's an [example](https://codesandbox.io/s/vuelidate-field-is-required-ggnsh?file=/src/App.vue:1543-1556) in CodeSandbox.

The problem is when to display the red * indicating that the field is required without passing extra props to the component in the library. In my example the star doesn't appear because there isn't enough information in the validation object to display it. The validation object is exactly the same whether the value is required or not.

The required field should look like this 
image


Furthermore, changing the prop of requiredIf(prop) doesn't reflect in the actual validation. Not sure if this is an another issue or if it's my mistake. The validator respects only he initial state. Here's the same [example](https://codesandbox.io/s/vuelidate-field-is-not-required-ezgr3?file=/src/App.vue) but the initial state is false. But that's besides the point.

dobromir-hristov (commented):
This implementation is to naive. If requiredIf changes, because the prop for falsy, I think the param you added will not update.

What you want is basically for a way, to know the outcome of the check, before the required method?

littlejon85:
Now this is getting interesting. Linked my branch on local to test whether it works and it does. The requiredIf(prop) validator checks only when prop = true as expected. In "next" (2.0.0-alpha.13) branch this doesn't work and only the initial status of prop is respected. I will investigate this further and probably open a new issue, if it doesn't exist already.

Anyway, the isRequired param gets propagated on prop change.

littlejon85:
Started issue regarding required* validators not respecting prop change https://github.com/vuelidate/vuelidate/issues/858

This PR accidentally solves that issue as a side effect but the issue should be resolved separately first.

dobromir-hristov:
The reason it may not work, is because we moved to watch instead of computed to trigger validators. We will most probably revert that though.

littlejon85:
> The reason it may not work, is because we moved to watch instead of computed to trigger validators. We will most probably revert that though.

I see. That makes sense. Since the isRequired param changes, the watcher updates the validation state.

> What you want is basically for a way, to know the outcome of the check, before the required method?

I'm not sure what you mean by " the outcome of the check, before the required method". The missing piece is the property which tells the requiredIf and requiredUnless validators to validate. I believe this should be part of the validation state for the reasons I tried to explain.

In vuelidate pre v2 this was kind of possible via $params in the root of validation state.
```js
$v.text  = { //vuelidate 0.x
...
    $params: {
        required: {
            type: "requiredIf", // this...
            prop: true   // ...together with this allowed to tell if the validator is active
        }
    }
}
```
It needed some hacky parsing though.

In v2 the validation state structure has changed and this is no longer available.

nakjemmy:
Thanks for pointing this out @littlejon85. I have been struggling with this issue.

Hopefully waiting for the PR merge.

dobromir-hristov (changes requested):
Sorry for the long wait, busy week

Looking at the way the old implementation was done, we were passing the `prop` directly to the params - https://github.com/vuelidate/vuelidate/blob/master/src/validators/requiredUnless.js

If we were to follow the exact convention as before, we would have to do:

```js
{
export default prop => ({
  $validator: requiredUnless(prop),
  $message,
  $params: { isRequired: prop }
})
```
But I dont know how that helped, because `prop` could be a string, that points to a property in the form.

But in Vuelidate 2, the `prop` can be a `ref`, `computed`, `boolean` or a `function`.

littlejon85:
Thanks for coming back Dobromir. We're all busy.

You're right, the fact that `prop` can be all these types makes it a little more complex in Vuelidate 2.

I spent some more time digging through the code trying to find an elegant way to achieve my intent without introducing unnecessary complexity.

I see 3 ways:

1. Abandon the `isRequired` param, pass through raw `prop` and let the users do whatever they need to do with it in their code. To follow the convention from before it would be necessary to add `$params.type` to all validators. 
```js
//...withMessages/requiredIf.js
export default function (prop) {
  return {
    $validator: requiredIf(prop),
    $message: 'The value is required',
 // $params: { isRequired: !!prop }
    $params: { 
      prop,
      type: 'requiredIf'
    }
  }
}
```
This would suffice for most cases but is not the most user friendly approach to for the use case I explained earlier.

2. Return the result of the predicate from `raw/required(If|Unless).js` and add it to `$params` as `isRequired`
```js
//raw/requiredIf.js
export default function requiredIf (propOrFunction) {
  return async (value, parentVM) => {
    const unwrappedProp = await unwrapPropOrFunction(this, propOrFunction, value, parentVM)
    return { $validator: validate(unwrappedProp, value), unwrappedProp }
  }
}

//raw/requiredUnless.js
export default function requiredUnless (propOrFunction) {
  return async (value, parentVM) => {
    const unwrappedProp = await unwrapPropOrFunction(this, propOrFunction, value, parentVM)
    return { $validator: validate(unwrappedProp, value), unwrappedProp }
  }
}
```

The function `unwrapPropOrFunction` is common for both conditionally required validators and would reside in `utils/common.js` 
```js
export async function unwrapPropOrFunction (context, propOrFunction, value, parentVM) {
  const unwrappedProp = typeof propOrFunction !== 'function'
    ? propOrFunction
    : await propOrFunction.call(context, value, parentVM)
  return unwrappedProp
}
```

The `$validator` and `unwrappedProp` would then be expanded in `withMessages/required(If|Unless).js`
```js
//withMessages/requiredIf.js
import requiredIf from '../raw/requiredIf'
export default function (prop) {
  const { $validator, unwrappedProp } = requiredIf(prop)
  return {
    $validator,
    $message: 'The value is required',
    $params: { isRequired: unwrappedProp }
  }
}

//withMessages/requiredUnless.js
import requiredUnless from '../raw/requiredUnless'
export default function (prop) {
  const { $validator, unwrappedProp } = requiredUnless(prop)
  return {
    $validator,
    $message: 'The value is required',
    $params: { isRequired: !unwrappedProp }
  }
}
```
This would let the user tell whether the predicate of the conditionally required validators resolved to `true` or `false`.

3. Combine 1 and 2
Provide both `isRequired` and `prop + type` in `$params` of `requiredIf` and `requiredUnless`. This would cover many unforeseen use cases and make it easy to solve the use case I explained earlier.

I'm okay with any of these solutions.

What are your thoughts?

dobromir-hristov:
Unfortunately that will not work. You will not have access to the inner function, until Vuelidate calls the validator.

```js
//raw/requiredIf.js
export default function requiredIf (propOrFunction) {
// this inner function, is called when Vuelidate runs the validator. It sets `this` to the component context, `value` and `parentVM` are passed as well there. At this point you cannot return `$validator`.
  return async (value, parentVM) => {
    const unwrappedProp = await unwrapPropOrFunction(this, propOrFunction, value, parentVM)
    return { $validator: validate(unwrappedProp, value), unwrappedProp }
  }
}
```

I think the only thing we can do, is pass the prop, that users pass, like Vuelidate 0.x, and the developer would have to handle it.

```js
export default function requiredIf(prop) {
 return withParams({ requiredIf: prop }, validatorFn)
}
```

Then you as a developer can do whatever you need to do
```js
const someRef = ref('')
const input = ref('')
const v = useVuelidate({ input: { required: requiredIf(someRef) } }, { input }) 
return { v, input, someRef }
```

```vue
<input :class="{ showStar: !!v.input.required.$props.requiredIf }" v-model="v.input.$model"
```

littlejon85:
You're right again. I agree. Passing through the prop will be enough.

However what about e.g. the `required` validator? In Vuelidate 0.x all validators included `type` in `$params` to identify the validator. The structure of validation state has changed in Vuelidate 2 and this was left out. I think this feature should remain.

To keep things consistent with the previous version I propose also returning the `type` to `$params` for all validators. The key name cannot be relied upon to programmatically identity the validators.

```js
//requiredIf.js
export default function (prop) {
  return {
    $validator: requiredIf(prop),
    $message: 'The value is required',
    $params: {
      type: 'requiredIf',
      prop
    }
  }
}

//required.js
export default {
  $validator: required,
  $message: 'Value is required',
  $params: { type: 'required' }
}

//same for alpha.js, alphaNum.js etc.
```

dobromir-hristov:
https://github.com/vuelidate/vuelidate/releases/tag/%40vuelidate%2Fvalidators%402.0.0-alpha.17 implements this.

westwick:
Hi @littlejon85 @dobromir-hristov , sorry to ping you guys, I'm facing a similar issue and struggling with how to achieve my result with this solution. My full issue is here if you have a chance: https://stackoverflow.com/questions/68657382/vue-vuelidate-dynamically-show-required-star-based-on-requiredif

But mostly wondering how to access the result of `requiredIf` in a child component. I see the validation function is now passed in to `$params.required.prop`, but not sure how to actually call it as it no longer has access to the right `this` context. Do I just need to bind it on the parent or something like that? Any code examples? Thanks for any help

jm-faqih:
Hi @westwick, I'm facing similar issue like you too right now. Did you find out the way to implement that?

PR #881, opened 2021-06-20
feature(core): pass more parameters to `$message` to help with i18n support

### LINKED ISSUE #769: Localizing error messages in Vuelidate-next
The application that I'm working on receives its localization strings dynamically from a network request. To localize strings, I have created a simple plugin, which receives a string name and refers to its value ( localizations are stored in Vuex ).

In short, its injected from a Nuxt plugin

```js
export default ({ store }, inject) {
  // locales = { homepage_title: "Homepage" }
  const t = path => store.state.locales[path]
  inject("t", t)
}
```

In templates, its used similarly to `$t('homepage_title')`.

How should I go about localizing my error messages with your library? I assume that I need to re-export the validators from your package that I'd like to use, and accustom their error messages to my use-case. 

Assuming I have this localization string for a `maxLength` validator `The maximum length allowed is {{length}}`, I would be able to do this

```js
export const maxLength = max => withMessage(
  vuelidateRule => $t('errors_maxlength', { length: validationRule.params.max }),
  maxLength(max)
)
```

My issue is, the `$t` part is a plugin injected at run-time, which I don't have access to in my example. How would I go about this to reduce the boilerplate for my forms to a minimum?

It feels like my end goal would be to be able to do this

```js
import { maxLength } from "@/utils/validators"

export default {
    ...,
    validations() {
        return {
            form: {
                name: {
                    maxLength: maxLength(35)
                }
            }
        }
    }
}
```

Thanks a lot for this feature, by the way. Now its a complete validator :)

dobromir-hristov:
## Summary

This PR adds a few extra properties to the `$message` function parameters, so its super easy to create simple and reusable translations on the client side. closes #769

We could probably have our own layer on top of `$message`, that ships with all messages in a file or something, so its easy for users to just provide their own instance of `$t` but I think the solution I provided in the docs is fine as is.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### DIFF
--- a/packages/docs/src/advanced_usage.md
+++ b/packages/docs/src/advanced_usage.md
@@ -411,9 +411,69 @@ export default {
 To clear out the external results, you can again, use the `$clearExternalResults()` method
 
 ```js
-async function validate() {
+async function validate () {
   this.$v.$clearExternalResults()
   // perform validations
   const result = await this.runAsyncValidators()
+}
+```
+
+## i18n support
+
+Validator messages are very flexible. You can wrap each validator with your own helper, that returns a translated error message, based on the
+validator name. Let's define a few validators:
+
+```js
+// @/utils/validators.js
+import { withI18nMessage } from '@/utils/withI18nMessage'
+import * as validators from '@vuelidate/validators'
+
+export const required = withI18nMessage(validators.required)
+export const minLength = withI18nMessage(validators.minLength)
+```
+
+Now lets see how that `withI18nMessage` helpers would look like:
+
+```js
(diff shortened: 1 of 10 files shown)


TijnvandenEijnde:
In my current project I use locale dependent error messages. Unfortunately once rendered they will not update when I switch locale. It looks to me that this PR might be able to solve this problem. I am very excited!

PR #887, opened 2021-07-06
Add support for function validators to `withI18nMessage`

wvffle:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
The `withI18nMessage` function proposed in the https://vuelidate-next.netlify.app/advanced_usage.html#i18n-support does not seem to work with function validators like `minLength`. I modified the function for it to be able to handle those validators well.

### DIFF
--- a/packages/docs/src/advanced_usage.md
+++ b/packages/docs/src/advanced_usage.md
@@ -603,11 +603,17 @@ import { i18n } from "@/i18n"
 
 const { t } = i18n.global || i18n // this should work for both Vue 2 and Vue 3 versions of vue-i18n
 
-export const withI18nMessage = (validator) => helpers.withMessage((props) => t(`messages.${props.$validator}`, {
-  model: props.$model,
-  property: props.$property,
-  ...props.$params
-}), validator)
+export const withI18nMessage = (validator, args = null) => {
+  if (validator instanceof Function && args === null) {
+    return (...args) => withI18nMessage(validator, args)
+  }
+
+  return validators.helpers.withMessage((props) => t(`messages.${props.$validator}`, {
+    model: props.$model,
+    property: props.$property,
+    ...props.$params
+  }), args === null ? validator : validator(...args))
+}
 ```
 
 We can now use the validators as we normally do
@@ -615,12 +621,12 @@ We can now use the validators as we normally do
 ```vue
 
 <script>
-import { required } from '@/utils/validators'
+import { required, minLength } from '@/utils/validators'
 
 export default {
   validations () {
     return {
-      name: { required }
+      name: { required, minLength: minLength(3) }
     }
   }
 }


dobromir-hristov:
Good catch, i just did that in 5 mins hehe. I will think about how to implement this with full compatibility.

We may need to bring this into the library directly, as it may become a bit more complicated. We need to figure out how users would provide their own translation instance.

wvffle:
Since `withMessage` takes function as first argument, `withI18nMessage` could take i18n instance instead.

dobromir-hristov:
> Since `withMessage` takes function as first argument, `withI18nMessage` could take i18n instance instead.

In most cases, you have a single i18n instance on your app, so I think it makes sense to have a factory function instead, that returns some "instance" of a `withI18nMessage`.

Something like this:

```js
import { createTranslatableMessages } from '@vuelidate/validators'
import i18n from '@/i18n'

export const withTranslation = createTranslatableMessages({ t: i18n.t })

// @/validators.js

export const required = withTranslation(validators.required)
export const minLength = withTranslation(validators.minLength)

// @/Form.vue

const validations = {
  name: {
    required,
    minLength: minLength(10)
  }
}
```

dobromir-hristov:
Looked into this, and the code below works great, I would have never thought to call the function on itself 😆 great job.

```js
import withMessage from './withMessage'

export default function createI18nMessage (t) {
  const withI18nMessage = (validator, args = null) => {
    if (typeof validator === 'function' && args === null) {
      return (...args) => withI18nMessage(validator, args)
    }
    return withMessage((props) => t(`messages.${props.$validator}`, {
      model: props.$model,
      property: props.$property,
      ...props.$params
    }), args === null ? validator : validator(...args))
  }
  return withI18nMessage
}
```

I am worried however about the TS users. I have been trying to create the types for this and I am hitting a wall...

wvffle:
I have never written any TS types, so I cannot help. Also if you want to push it as a feature, I'd rewrite it as:

```js
import withMessage from './withMessage'

export default function createI18nMessage (t) {
  const translate = (props) => t(`messages.${props.$validator}`, {
    model: props.$model,
    property: props.$property,
    ...props.$params
  })
    
  const withI18nMessage = (validator) => {
    if (typeof validator === 'function') {
      return (...args) => withMessage(translate, validator(...args))
    }
    
    return withMessage(translate, validator)
  }
  
  return withI18nMessage
}
```

dobromir-hristov:
That i think wont work if you just pass a simple validator as a function, right?

```js
const isTrurthy = (value)=> !!value

withI18n(isTrurthy)

```

dobromir-hristov:
Check #894, let me know what you think.

dobromir-hristov:
I am closing this PR, #894 should cover it.

PR #903, opened 2021-07-16
feat(types): added TS cast support to `or`, `not`, `and`. Added vm params in `ValidatorFn`

victorgarciaesgi:
Added TS cast support and vm params

- [x] Feature

- [ ] Yes
- [x] No

Fellowing previous PR, I added more Typescript upgrades for validators.

- `and`: support for type casting
- `or`: support for type casting
- `not`: support for type casting
- `ValidatorFn`: added `vm: Vue` as second param to access current instance

### DIFF
--- a/packages/vuelidate/index.d.ts
+++ b/packages/vuelidate/index.d.ts
@@ -40,7 +40,7 @@ export interface ValidatorResponse {
   [key: string]: any
 }
 
-export type ValidatorFn <T = unknown> = (value: T) => boolean | ValidatorResponse | Promise<boolean | ValidatorResponse>;
+export type ValidatorFn <T = unknown> = (value: T, vm: Vue) => boolean | ValidatorResponse | Promise<boolean | ValidatorResponse>;
 
 export interface ValidationRuleWithoutParams <T = unknown> {
   $validator: ValidatorFn<T>
--- a/packages/validators/index.d.ts
+++ b/packages/validators/index.d.ts
@@ -9,8 +9,8 @@ import { Ref } from 'vue-demi';
 // Rules
 export const alpha: ValidationRuleWithoutParams;
 export const alphaNum: ValidationRuleWithoutParams;
-export const and: (
-  ...validators: ValidationRule[]
+export const and: <T = unknown>(
+  ...validators: ValidationRule<T>[]
 ) => ValidationRuleWithoutParams;
 export const between: (
   min: number | Ref<number>,
@@ -33,10 +33,10 @@ export const minLength: (
 export const minValue: (
   min: number | Ref<number> | string | Ref<string>
 ) => ValidationRuleWithParams<{ min: number }>;
-export const not: (validator: ValidationRule) => ValidationRuleWithoutParams;
+export const not: <T = unknown>(validator: ValidationRule<T>) => ValidationRuleWithoutParams;
 export const numeric: ValidationRuleWithoutParams;
-export const or: (
-  ...validators: ValidationRule[]
+export const or: <T = unknown>(
+  ...validators: ValidationRule<T>[]
 ) => ValidationRuleWithoutParams;
 export const required: ValidationRuleWithoutParams;
 export const requiredIf: (prop: boolean | string | (() => boolean | Promise<boolean>)) => ValidationRuleWithoutParams;


dobromir-hristov:
Hey, sorry for the long wait. I just merged #894 which changed a bit the validators types, would you mind resolving the conflict and maybe looking over the new types I added? I am sure those can be improved a bit.

victorgarciaesgi:
Hi @dobromir-hristov no problem I will look into this when I have time :)

packages/vuelidate/index.d.ts
```diff
@@ -40,7 +40,7 @@ export interface ValidatorResponse {
   [key: string]: any
 }
 
-export type ValidatorFn <T = unknown> = (value: T) => boolean | ValidatorResponse | Promise<boolean | ValidatorResponse>;
+export type ValidatorFn <T = unknown> = (value: T, vm: Vue) => boolean | ValidatorResponse | Promise<boolean | ValidatorResponse>;
```
subvertallchris, line 43: Is this always correct? I'm checking this in a Vue 2.6 app using the Composition API library and the type of `vm` appears to be `Vue & ThingsAvailableToTheTemplate`, where `ThingsAvailableToTheTemplate` is dependent on the component itself. `Vue` by itself doesn't have the appropriate scope.

subvertallchris:
This is a blocker for me. The `ValidatorFn` type in general seems kind of broken, too. When trying to create a custom validator function, its generic param allows me to define the rule but when I pass an object to `useVuelidate`, it complains about `unknown`. I think this is because of this:

```typescript
export interface ValidationArgs {
  [K: string]: ValidationRule | ValidationArgs
}
```

I think that's going to default back to `unknown` so an earlier definition won't matter. Haven't dug into it too much, maybe I'm doing something wrong?

packages/vuelidate/index.d.ts
dobromir-hristov: the second parameter is the component's Vue instance yeah.

subvertallchris: Yes, but the type wouldn’t be `Vue` since that gives you something more general, the TypeScript compiler doesn’t know that it represents this actual component. I haven’t come across a type in Vue that does this for you, I think you need a second generic argument, something like `V extends Vue = Vue`, that will allow the user to describe it. Even that might be too restrictive, so something like `V = any` will at least allow the user to handle it while a more appropriate type is figured out.

dobromir-hristov: ahaa, well I am still a TS noob :/ Would you care making a PR so we can clean such things up?

subvertallchris:
I’ll give it a shot this week. I’m using this in a project at work so I can
probably get some time for it on the clock.

packages/vuelidate/index.d.ts
basti-n: what about specifying the second param as a generic which extends the `Vue` component
`export type ValidatorFn <T = unknown, K extends Vue> = (value: T, vm: K) => boolean | ValidatorResponse | Promise<boolean | ValidatorResponse>;
`

subvertallchris: @basti-n I think that's the winning move since it would let the user just feed the function the things they know that the compiler couldn't know otherwise.

I think the use of `unknown` here is a problem, too. This fails to compile:

```typescript
  setup(props) {
    const specialRule: ValidatorFn<string | null> = (input) => {
      const confidentAboutInput = input as string | null;
      return !!confidentAboutInput && confidentAboutInput !== 'foo';
    };

    const rules = {
      target: {
        value: {
          specialRule
        }
      }
    };

    const vvv = useVuelidate(rules, props);
  }
```

B6C57E7E-5D12-4E4C-A0EB-C7F8F8BE41D2

The definition of my special rule is fine but `useValidate` is failing. This happens because `useVuelidate` has these types for its first parameter:

```typescript
validationsArgs: Ref<Vargs> | ValidationArgs,

# which refers to...

export interface ValidationArgs {
  [K: string]: ValidationRule | ValidationArgs
}

# which refers to...

export type ValidationRule <T = unknown> = ValidationRuleWithParams<any, T> | ValidationRuleWithoutParams<T> | ValidatorFn<T>;

# which refers to...

export type ValidatorFn <T = unknown> = (value: T) => boolean | ValidatorResponse | Promise<boolean | ValidatorResponse>;
```

Since `ValidationArgs` is passed without a generic no matter what the user does when calling `useVuelidate`, it means that it's going to compare my deliberately typed `specialRule: ValidatorFn<string | null>` to `ValidatorFn<unknown>`. I'd probably swap those `T = unknown` for `T = any` (or would `T extends unknown` do the trick?) to at least let users take control over it and avoid this compiler error. Alternatively, you could write a type that extracts the value of T from any `ValidatorFn`s passed into `validationsArgs`, but I'm not sure what that would take.

@dobromir-hristov, I wasn't able to get a good dev environment working for this library with TypeScript. If anyone's got an hour free sometime soon and wants to do some remote pairing on it, I'm happy to jump in.

basti-n: Having the value of T extracted certainly is better from a type-correctness PoV. For the time being, and given that typing with `any` seems not to be prohibited by linting going with `any` c/would also be fine. 

For your case, why don't you keep the `ValidatorFn<unknown>` and have the type narrowed to string or null within the specialRule? 

```
const isStringOrNull = (val: unknown): val is number | null => {
    return typeof val === 'string' || val === null
}
```

subvertallchris: We're doing something like that as a workaround but it seems like the intention is for `ValidatorFn<string | unknown>` to let us avoid doing that. I think use of `any` + your `Record` example above would be a good solution to unblock folks for now while more permanent types are explored.

PR #927, opened 2021-09-01
feat(core): allow passing configs to OptionsApi directly in `useVuelidate`, closes #922

### LINKED ISSUE #922: Can't have a ref for external results when Options API is used
**Describe the bug**

Vuelidate accesses `vuelidateExternalResults` by value which makes it impossible to define external results as a `ref` when the Options API is used. In the example below,  the `fail` won't cause any effect because the `vuelidateExternalResults` will be read by value once and so the following updates to the prop won't be noticed.

```js
export default {
  setup: () => ({
    v$: useVuelidate(),
    vuelidateExternalResults: ref(null)
  }),
  data: () => ({
    myField: null
  }),
  methods: {
    fail() {
      this.vuelidateExternalResults = { myField: 'Server is down' }
    }
  } 
}
```

An obvious fix is to use `toRef` but I saw some complex logic inside `$clearExternalResults`, so it might not work or require the preservation of the `vuelidateExternalResults`'s type. Here is the place where the `vuelidateExternalResults` gets accessed by value and an imaginary fix for the problem:

https://github.com/vuelidate/vuelidate/blob/64892a5ebcdee1ad8c728766d9e693f605e36477/packages/vuelidate/src/index.js#L153

```js
setValidations({
  ...otherOptions,
  externalResults: toRef(instance.proxy, 'vuelidateExternalResults'),
  areExternalResultsReactive: isReactive(instance.proxy.vuelidateExternalResults) // for the later use in `$clearExternalResults`
})
```

**Reproduction URL**

https://codesandbox.io/s/dreamy-spence-plni5?file=/src/App.vue

**To Reproduce**

The demo automatically sets the `vuelidateExternalResults` and makes the `number` dirty on init, so no actions are required.

**Expected behavior**

The `Server is down` is expected to be shown but it's missing because Vuelidate didn't notice the replacement of the `vuelidateExternalResults`. Check the `fail` method for details.

dobromir-hristov:
## Summary

This PR allows for `useVuelidate` to accept a config object, when used in the OptionsAPI, closes #922 

This should aid in providing reactive data, like `$externalResults` more easily.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/packages/vuelidate/src/index.js
+++ b/packages/vuelidate/src/index.js
@@ -75,6 +75,8 @@ function nestedValidations ({ $scope }) {
  * @property {String | Number | Symbol} [$scope] - A scope to limit child component registration
  * @property {Boolean} [$stopPropagation] - Tells a Vue component to stop sending its results up to the parent
  * @property {Ref<Object>} [$externalResults] - External error messages, like from server validation.
+ * @property {Boolean} [$autoDirty] - Should the form watch for state changed, and automatically set `$dirty` to true.
+ * @property {Boolean} [$lazy] - Should the validations be lazy, and run only after they are dirty
  */
 
 /**
@@ -150,12 +152,12 @@ export function useVuelidate (validations, state, globalConfig = {}) {
             resultsCache,
             globalConfig,
             instance: instance.proxy,
-            externalResults: instance.proxy.vuelidateExternalResults
+            externalResults: $externalResults || instance.proxy.vuelidateExternalResults
           })
         }, { immediate: true })
     })
 
-    globalConfig = componentOptions.validationsConfig || {}
+    globalConfig = componentOptions.validationsConfig || globalConfig
   } else {
     const validationsWatchTarget = isRef(validations) || isProxy(validations)
       ? validations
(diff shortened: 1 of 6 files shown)


packages/vuelidate/src/index.js
```diff
@@ -150,12 +152,12 @@ export function useVuelidate (validations, state, globalConfig = {}) {
             resultsCache,
             globalConfig,
             instance: instance.proxy,
-            externalResults: instance.proxy.vuelidateExternalResults
+            externalResults: $externalResults || instance.proxy.vuelidateExternalResults
```
nskazki, line 155: Thanks for doing it! I really like how responsive you are to the needs of community!

dobromir-hristov: Note this will still not work for Vue 2, without pre-defining the keys. This is a Vue 2 limitation of the reactivity API.

nskazki: I understand, but I can start with `ref(null)`, return `$externalResults`  from the `setup` hook, and then replace that `null` with any other value and the reference will reflect it, like so `this.$externalResults = serverResponse`.

PR #941, opened 2021-09-22
Types for or/not/and, `ValidatorFn`, relax types (PR #903 + mods)

### LINKED ISSUE #917: Types for custom validator with vm missing in ValidationRule
**Describe the bug**
I wrote a custom validator using the second parameter vm of the Validator Function [like described here](https://vuelidate-next.netlify.app/custom_validators.html#passing-extra-properties-to-validators). 
I also have a withI18nMessage function to pass all validators thru to add messages. This function uses the `helpers.withMessage`-Function. This function requires 2 parameters with the following types `message: string | Function, validator: ValidationRule`. Unfortunately I can't pass my validator with type `(value: T, vm: any) => boolean | ValidatorResponse | Promise<boolean | ValidatorResponse>` to the validator parameter, because the vm option is not included in ValidationRule.

Should I use a different type or could you please just add it?

**Expected behavior**
I should be able to use `helper.withMessage` with custom validators with the vm parameter too.

subvertallchris:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

I do not expect this to be a breaking change for TypeScript users but I think that there is still room for improvement.

**Other information:**

This is a fork of #903 which appears to resolve #917 with the changes discussed in the original PR's comments. The goal of this modded version of the original PR is to unblock teams like mine that need typed `vm` and are unable to use explicitly typed `ValidatorFn`.

I still have not figured out a good dev setup for this library so I unfortunately have not tested it; however, I do not expect any of the changes to be problematic as all they're doing is adding some optional types and adding some generics.

The original PR by @victorgarciaesgi says:

```
Fellowing previous PR, I added more Typescript upgrades for validators.

and: support for type casting
or: support for type casting
not: support for type casting
ValidatorFn: added vm: Vue as second param to access current instance
```

In the comments, we agreed that `ValidatorFn`'s `vm` should be typed as the union between _some component-specific state_ and `Vue`, which is one of the changes made here. The other changes, swapping out `unknown` for `any`, works around an issue wherein the library has types that reference other types which default to `unknown`, effectively preventing users from declaring `ValidatorFn` as described in [this comment here](https://github.com/vuelidate/vuelidate/pull/903#discussion_r710268529). I think that `unknown` is probably the correct type for `ValidatorFn` but if and only if `useVuelidate` is able to extract appropriate types and feed them down the pipe internally. I'd be happy to try this if someone can help me get a good TypeScript dev flow working.

@victorgarciaesgi, I hope I did not overstep by jumping on this, thank you for your work. 🙏  @basti-n maybe you can eyeball this and weigh in?

### DIFF
--- a/packages/validators/index.d.ts
+++ b/packages/validators/index.d.ts
@@ -47,8 +47,8 @@ export const sameAs: <E = unknown>(
 ) => ValidationRuleWithParams<{ equalTo: E, otherName: string }>;
 export const url: ValidationRuleWithoutParams;
 export const helpers: {
-  withParams: (params: object, validator: ValidationRule) => ValidationRuleWithParams
-  withMessage: (message: string | Function, validator: ValidationRule) => ValidationRuleWithParams
+  withParams: <T = unknown>(params: object, validator: ValidationRule<T>) => ValidationRuleWithParams
+  withMessage: <T = unknown>(message: string | ((params: MessageProps) => string), validator: ValidationRule<T>) => ValidationRuleWithParams
   req: Function
   len: Function
   regex: Function
--- a/packages/validators/index.d.ts
+++ b/packages/validators/index.d.ts
@@ -3,14 +3,14 @@ import {
   ValidationRuleWithParams,
   ValidationRule,
   ValidationArgs
-} from '@vuelidate/core'
+} from '@vuelidate/core';
 import { Ref } from 'vue-demi';
 
 // Rules
 export const alpha: ValidationRuleWithoutParams;
 export const alphaNum: ValidationRuleWithoutParams;
-export const and: (
-  ...validators: ValidationRule[]
+export const and: <T = unknown>(
+  ...validators: ValidationRule<T>[]
 ) => ValidationRuleWithoutParams;
 export const between: (
   min: number | Ref<number>,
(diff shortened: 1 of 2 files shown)


dobromir-hristov:
@subvertallchris thank you for your contribution. 

As an idea on how to test this, just run `npm run build` and then `npm pack` inside the `packages/vuelidate` folder.

Then you should be able to just install that into your test project using `npm install ../path/to/tar`.

subvertallchris:
@dobromir-hristov This is very helpful, thank you. I was able to confirm that it resolved my immediate issues around `ValidatorFn`, the type of `vm` and the issue caused by the default `unknown` generic param that would get passed around after calling `useVuelidate`. I'll have to poke around our app to see if there are components I can refactor quickly with rules that use `and`, `or, or `not`.

I'd also like to get another TypeScripter's opinion on the move from `unknown` to `any` since it's usually a bad idea, maybe someone has a better suggestion for the short-term.

subvertallchris:
@dobromir-hristov Any thoughts on this? I know I said I'd like another TypeScripter's opinion but if nobody volunteers, IMO it's worth moving ahead. They'll certainly make themselves known if there's a problem. 😈

packages/validators/index.d.ts
```diff
@@ -32,23 +32,23 @@ export const minLength: (
...
 export const sameAs: <E = unknown>(
   equalTo: E,
   otherName?: string
 ) => ValidationRuleWithParams<{ equalTo: E, otherName: string }>;
 export const url: ValidationRuleWithoutParams;
 export const helpers: {
-  withParams: <T = unknown>(params: object, validator: ValidationRule<T>) => ValidationRuleWithParams
-  withMessage: <T = unknown>(message: string | ((params: MessageProps) => string), validator: ValidationRule<T>) => ValidationRuleWithParams
+  withParams: (params: object, validator: ValidationRule) => ValidationRuleWithParams
+  withMessage: (message: string | Function, validator: ValidationRule) => ValidationRuleWithParams
```
dobromir-hristov, line 51: Why are we removing the more detailed declaration for this? Is it better to just have a function?

dobromir-hristov (commented):
Overall I think its fine. If no one else has any complaints, I can just merge it.

subvertallchris:
Cool, thanks. Not sure about your one question, can’t remember if I did
that or if that was in the original. I can check it with more context and
let you know on Monday. Thanks for checking in with this!

packages/validators/index.d.ts
subvertallchris: This change was part of the original PR, I think it was done before you updated the types in the main branch. I'm not sure what the intention was. I can't think of a time that I've ever used the `Function` keyword like that and don't want to speculate so I'm just going to update these to match main.

subvertallchris:
@dobromir-hristov I just modified the PR to fix those two function signatures that you mentioned in a comment. I rebased and forced push to my branch so your comment was wiped out but you should have received a github notification before that happened.

dobromir-hristov (approved):

subvertallchris:
Thank you! 🙌 

Would you mind releasing a new alpha build?

dobromir-hristov:
Will do tonight yeah

PR #951, opened 2021-10-12
feat(core): add reward early mode, closes  #897

### LINKED ISSUE #897: API to support: reward early, punish late UX
**Is your feature request related to a problem? Please describe.**
The best UX when it comes to validation is a concept sometimes referred to as "reward early, punish late".

In practice, it means that:
- if a value is invalid, as soon as the user fixes the input it should be green again _(reward early)_.
- if a value is valid, it shouldn't become red until the user finishes its input  -- typically at `blur` or `change` events _(punish late)_.

A sample interaction, would be:
- Say field A must be at least 4 characters, and is currently "1235" (valid).
- User wants to fix this to 1234, so it removes the "5" -> the input should not flash red yet!
- User types the "4"
- Value is committed (e.g. blur event), 1234 is valid, field stays green.

On the other hand, if field was "123" and red, as soon as the user types "4" the error should go away.

I have built such UX on top of vuelidate, but it could easily have built-in support, with just two small additions that I'm proposing below.

**Describe the solution you'd like**
Today `$error` indicates if the error should be displayed in UI, and is based on `$invalid` + `$touch()` events (a non-dirty value is not displayed in red).

I propose to optionally change `$error` behavior to match the UX described above.
This would be opted-in by users using with a new option `{ useCommit: true }`.

There would be a single new API `$commit()` to support this feature.

How it would work: 
- If option `{ useCommit: true }` is not used -> everything works as today. `$commit()` is a synonym for `$touch()`.

Otherwise: 
- When touched, if `$invalid` is false then update `$error = false`, like today.
- When touched, if `$invalid` becomes true -> do nothing!
- When `$commit()` is invoked, call `$touch()`, then set `$error = $invalid`.

And that's it.

**Describe alternatives you've considered**
You can build this on top of the Vuelidate object, but it's a bit hacky / more complicated.

dobromir-hristov:
## Summary

Adds the reward-early-punish-late mode to Vuelidate. Closes  #897

## How to use

You just set a `$rewardEarly: true` option to your Vuelidate instance.

Then you can call `$commit` or `$validate` to re-run validators. 

```vue

<template>
  <div>
    <input v-model="v$.name.$model" @blur="v$.name.$commit" />
    <p v-for="error of v.$errors" :key="error.$uid">
      {{ error.$message }}
    </p>
  </div>
</template>
<script>
import useVuelidate from '@vuelidate/core'
import { required } from '@vuelidate/validators'

export default {
  setup () {
    const rules = {
      name: { required }
    }
    const name = ref(null)
    return {
      v$: useVuelidate(rules, { name }, { $rewardEarly: true }),
    }
  }
}
</script>
```

https://user-images.githubusercontent.com/9863944/137030230-8dbd8fc0-9c2a-4b87-a9d4-71bcde1adf28.mov

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/packages/docs/src/index.md
+++ b/packages/docs/src/index.md
@@ -125,19 +125,19 @@ export default {
 }
 ```
 
-Now that validations are set up, we can check inside our template for errors by looking at the `name` property inside of the `v$` Vuelidate object. It will hold all the information and state of our `name` state's validation.
+Now that validations are set up, we can check inside our template for errors by looking for example at the `firstName` property inside of the `v$` Vuelidate object. It will hold all the information and state of our `firstName` state's validation.
 
-If _any_ error is present, the `$errors` array property inside of `$v.name` will contain an object that describes each error for us to loop through.
+If _any_ error is present, the `$errors` array property inside of `$v.firstName` will contain an object that describes each error for us to loop through.
 
 Each object inside the `$errors` array will contain a few properties that allows us to dynamically build our error message.
 
-An example of our `name` property being in an error state due to it being `required` would be:
+An example of our `firstName` property being in an error state due to it being `required` would be:
 
 ```js
 {
-  "$property": "name",
+  "$property": "firstName",
   "$validator": "required",
-  "$message": "The value is required",
+  "$message": "Value is required",
   [...]
 }
 ```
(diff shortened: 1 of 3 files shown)


JohnCampionJr:
I would like to but I cannot get the test project to run in the dev code.   I feel like it must be something I'm doing wrong, but it will not start. It seems to be missing the references to vuelidate itself.

JohnCampionJr:
Now it works perfectly; this is awesome, thanks!!

JohnCampionJr:
Any chance this gets published soon? 👍

PR #977, opened 2021-11-20
docs: fix typos in `custom_validators.md`

chalkygames123:
If my understanding is correct, these should be typos.


chalkygames123:
If my understanding is correct, these should be typos.

dobromir-hristov (approved):
Yep, thanks

PR #1024, opened 2022-03-27
fix build issue with typescript undefined error

A-Ghorab:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

fix build issue with typescript undefined error

i redid some checks on the previous pull request i did today and just found that during the build process it marks v$.field.$error as undefined so to force the ValidationArgs to have a value i just removed the optional part from the index.d.ts

### DIFF
--- a/packages/vuelidate/index.d.ts
+++ b/packages/vuelidate/index.d.ts
@@ -58,7 +58,7 @@ export type ValidationRule <T = any> = ValidationRuleWithParams<any, T> | Valida
 export type ValidationRuleCollection <T = any> = Record<string, ValidationRule<T>>;
 
 export type ValidationArgs<T = unknown> = {
-  [key in keyof T]?: ValidationArgs<T[key]> | ValidationRuleCollection<typeof T[key]> | ValidationRule<typeof T[key]>
+  [key in keyof T]: ValidationArgs<T[key]> | ValidationRuleCollection<typeof T[key]> | ValidationRule<typeof T[key]>
 }
 
 export interface RuleResultWithoutParams {
--- a/packages/vuelidate/index.d.ts
+++ b/packages/vuelidate/index.d.ts
@@ -58,7 +58,7 @@ export type ValidationRule <T = any> = ValidationRuleWithParams<any, T> | Valida
 export type ValidationRuleCollection <T = any> = Record<string, ValidationRule<T>>;
 
 export type ValidationArgs<T = unknown> = {
-  [key: string]: ValidationArgs<T[key]> | ValidationRuleCollection<typeof T[key]> | ValidationRule<typeof T[key]>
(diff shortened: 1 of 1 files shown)


dobromir-hristov:
Can you give me a bit more details on what this fixes? What was optional, and now isn't?

A-Ghorab:
> Can you give me a bit more details on what this fixes? What was optional, and now isn't?

Sure sorry that i didn't elaborate on the issue

currently one of the issues with typescript is that if do this 

```ts
const v$ = useVuelidate(rules,state);
````

it forces that i should specify every item in the state and set a rule for it even if it don't need to.
so in the previous pull request 
i made the typescript look like this
```ts
export type ValidationArgs<T = unknown> = {
  [key in keyof T]?: ValidationArgs<T[key]> | ValidationRuleCollection<typeof T[key]> | ValidationRule<typeof T[key]>
}
```

the "?" means every key in the Type can be set (optional) and it's not required which does match what this package does

but during my tests with the new version
in the template section 
```ts
:invalid="v$.field.$error"
```
it was working in the development without any errors but when i made a production build 
it gave me an error that field might be undefined because typescript mark it as an optional so it might not have a value and this isn't correct 

so we need to remove the optional and make it required for now until typescript have a better way to mark the field as optional but not undefined and exist in the state object

hope the description of the issue is now better

dobromir-hristov:
But can you still leave state Fields without validation rules? Not everything is required to validate.

A-Ghorab:
maybe this weekend i will revise the typescript definition and come back with a better solution for all of this 😄

A-Ghorab:
> But can you still leave state Fields without validation rules? Not everything is required to validate.

yes right now for ts i can trick it with this simple trick 

```ts
const v$ = useVuelidate(rules,state as **any**);
```
which removes the errors without any issues.

i will try to revise the typescript definition file one more time maybe i can find a smarter solution that works well on both cases 😄

A-Ghorab:
i made a little change to force the state to include all validation fields but not the other way around 

so Validation can have like 1 field to validate and it will make sure that this field exist in the State object

AGhorabHekouky:
Right now I'm doing some testing in my app to make sure its 100% working without issues

A-Ghorab:
Right now I'm doing some testing in my app to make sure its 100% working without issues

A-Ghorab:
@dobromir-hristov i tested this new syntax and it's going through every scenario


- [x] State has field that won't be validated, Rule will be accepted without an issue
- [x] Rules has a field that doesn't exist in the State (will give an error)
- [x] Rule Result is showing correctly without error in the Template like v$.field.$error

dobromir-hristov (approved):

dobromir-hristov:
Lets try :D

PR #1082, opened 2022-07-17
fix  #1079 - Add support for Vue 2.7

### LINKED ISSUE #1079: Please update vue-demi dependency to 0.13.x for Vue 2.7 support
Vue 2.7 have composition API integrated in the package, but vuelidate require @vue/composition-api, now i need to keep both for properly work and have some side effects from that.

Florent-Bouisset:
## Summary

Add support for vue 2.7

Allow to unit test for 2.7 with vue-demi

fixes #1079

Even if vue 2.7 doesn't need `vue-template-compiler` itself
it's still needed because of `@vue/test-utils`  and need to match the current vue version, so it is why I added this behavior to `swap-vue.js`

https://blog.vuejs.org/posts/vue-2-7-naruto.html
> Upgrade vue to ^2.7.0. You can also remove vue-template-compiler from the dependencies - it is no longer needed in 2.7.
> 
> Note: if you are using @vue/test-utils, you may need to keep it in the dependencies for now, but this requirement will also be lifted in a new release of test utils.

fixes #1079

## Metadata

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

**Other information:**

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,8 @@
     "standard-version": "^9.3.1",
     "vue": "^3.2.2",
     "vue-template-compiler": "^2.6.14",
-    "vue2": "npm:vue@2.6.14"
+    "vue2": "npm:vue@2.6.14",
+    "vue2.7": "npm:vue@2.7.5"
   },
   "scripts": {
     "lerna:publish": "lerna publish --conventional-commits",
@@ -60,9 +61,11 @@
     "release:dry": "yarn lerna:build && yarn lerna:version --no-push --no-git-tag-version",
     "test:unit": "npm run test:3",
     "test:2": "npm run use-vue:2 && jest",
+    "test:2.7": "npm run use-vue:2.7 && jest",
     "test:3": "npm run use-vue:3 && jest",
-    "test:all": "npm run test:2 && npm run test:3",
+    "test:all": "npm run test:2 && npm run test:2.7 && npm run test:3",
     "use-vue:2": "node scripts/swap-vue.js 2 && vue-demi-switch 2",
+    "use-vue:2.7": "node scripts/swap-vue.js 2.7 && vue-demi-switch 2.7",
     "use-vue:3": "node scripts/swap-vue.js 3 && vue-demi-switch 3"
   },
   "config": {
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "url": "https://github.com/vuelidate/vuelidate.git"
   },
   "dependencies": {
-    "vue-demi": "^0.12.0"
+    "vue-demi": "^0.13.4"
   },
   "optionalDependencies": {
     "@vue/composition-api": "^1.3.1"
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "standard-version": "^9.3.1",
     "vue": "^3.2.2",
     "vue-template-compiler": "^2.6.14",
+    "vue-template-compiler2.7": "npm:vue-template-compiler@2.7.5",
     "vue2": "npm:vue@2.6.14",
     "vue2.7": "npm:vue@2.7.5"
   },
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -13,7 +13,7 @@
     "url": "https://github.com/vuelidate/vuelidate.git"
   },
   "dependencies": {
-    "vue-demi": "^0.12.0"
+    "vue-demi": "^0.13.4"
   },
   "optionalDependencies": {
     "@vue/composition-api": "^1.3.1"
```
dobromir-hristov, line 19: Should we update the composition api version too?

packages/vuelidate/src/index.js
```diff
@@ -55,7 +55,7 @@ export function useVuelidate (validations, state, globalConfig = {}) {
     sendValidationResultsToParent,
     removeValidationResultsFromParent
   } = instance
-    ? nestedValidations({ $scope, instance })
+    ? nestedValidations({ $scope, instance: instance.proxy })
```
dobromir-hristov: You sure this is compatible with 2 2.7 and 3 ? I remember I had issues with the `getCurrentInstance` function a few times

scripts/swap-vue.js
```diff
@@ -2,8 +2,15 @@ const fs = require('fs')
 const path = require('path')
 
 const Vue2 = path.join(__dirname, '../node_modules/vue2')
+// eslint-disable-next-line camelcase
```
dobromir-hristov: you can disable it for the entire file :)

dobromir-hristov (commented):
This looks great, thank you so much! I saw all the tests have passed in CI, just checking to see if it works in real world too.

I am having some issues with my local, so I cant test it right now.

packages/vuelidate/src/index.js
dobromir-hristov: Actually why do we even need to do this? We just attach something to the instance, not sure we need to pass the `ComponentPublicInstance`

Florent-Bouisset: I made test with a real app only for 2.7, I trusted the units test for 2.6 and 3 but I'm going to try it

the function `getCurrentInstance` has different signature with 2.6 / 2.7 / 3

// 2.6  unofficial plugin @vue/composition-api
```js
declare function getCurrentInstance(): ComponentInternalInstance | null;
```

// 2.7
```js
export function getCurrentInstance(): { proxy: Vue } | null 
```

// 3
```js
export declare const getCurrentInstance: () => ComponentInternalInstance | null;
```

According to what has been said in this discussion:
https://github.com/vuejs/vue/issues/12596#issuecomment-1173269807

If I understand it right, we should NOT use the internal Instance but the public instance, which can be accessed by  `getCurrentInstance().proxy`

Maybe I'm wrong let me know what you understood about it

dobromir-hristov: Great find, let's keep it that way then. I mainly was not sure if vue 2 would have proxy, so was worried not to regress.

package.json
Florent-Bouisset: I don't think it's necessary ?

ademyalcin27:
Is there any update here?, We are waiting for this development?

dobromir-hristov:
I think we are good. I will test this Saturday and cut a release :) thank you so for the patience

dobromir-hristov (approved):
amazing work :)

packages/vuelidate/src/index.js
robertmoura: I'm running Vue 3 and proxy doesn't have a `uid` on it. Means that the validations are registered as `_vuelidate_undefined`. Causing issues with multiple child validations.
```
$registerAs = `_vuelidate_${uid}`
```

Florent-Bouisset: Sorry for that :( 

uid is not on the `ComponentPublicInstance` but on the  `ComponentInternalInstance` on vue3 

we can do as before by accessing the internal instance with `instance.$.uid`
```js
    const uid = instance?.$?.uid || instance._uid
    $registerAs = `_vuelidate_${uid}`
```

but as said before accessing internal instance is not recommended, 
maybe we should create our own uid ? I don't think we need to have the same as the component one.

any thoughts @dobromir-hristov ?

PR #1228, opened 2024-02-26
fix: validation of missing properties in `helpers.forEach`

haug1:
## Summary

_Describe what your PR does in a few short words_

validations on properties inside `helpers.forEach` that were missing in the target model were skipped

(I think there are several issues referencing this... I'm just fixing it because I need to deal with it in an app that I'm maintaining. Please get this merged so I can unpin my modified version of `vuelidate` from my repo.)

Here's a simple reproduction:

```vue
<template>
  <div
    v-for="(input, index) in state.collection"
    :key="index"
  >
    <input v-model="input.name" type="text" />
    <div
      v-for="error in v$.collection.$each.$response.$errors[index].name"
      :key="error"
    >
      {{ error.$message }}
    </div>
  </div>
</template>
<script>
import { helpers, required } from "@vuelidate/validators";
import { useVuelidate } from "@vuelidate/core";
import { reactive } from "vue";

export default {
  setup() {
    const state = reactive({
      collection: [{ name: "" }, {}],
    });
    return {
      v$: useVuelidate(
        {
          collection: {
            $each: helpers.forEach({
              name: {
                required,
              },
            }),
          },
        },
        state
      ),
      state,
    };
  },
};
</script>
```

Running the above code, you will notice that `vuelidate` incorrectly does not report errors for the second object in the array.

## Metadata

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No
- [X] Maybe?

The validation result does change a tiny bit, see the diff in the test. Properties without validators are now omitted from the validation result instead of appearing empty. I assume this change may possibly break a consumer app with very specific implementations of how they read the validation result from the `forEach`. I've decided to ignore this possibly breaking change, since it seems inherently counter intuitive that properties without validators appear in the validation result and that is different behavior to how it normally behaves outside of the `forEach` function. Any implementation that leans on this behavior must be doing it for no particular good reason, only a quirk they've had to accomodate for in the past.

### DIFF
--- a/packages/validators/src/utils/__tests__/forEach.spec.js
+++ b/packages/validators/src/utils/__tests__/forEach.spec.js
@@ -49,8 +49,7 @@ describe('forEach', () => {
             required: false,
             $invalid: true,
             $error: true
-          },
-          surname: {}
+          }
         }
       ],
       $errors: [
@@ -74,8 +73,7 @@ describe('forEach', () => {
               $response: false,
               $validator: 'required'
             }
-          ],
-          surname: []
+          ]
         }
       ],
       $valid: false
@@ -313,5 +311,11 @@ describe('forEach', () => {
     expect(forEach(rules).$validator([
       { name: 'Foo' }, { name: 'Foo' }
     ]).$valid).toBe(true)
+  })
+
+  it('returns invalid when required property is missing', () => {
+    const validate = forEach({ something: { required } }).$validator
+    expect(validate([{}]).$valid).toBe(false)
+    expect(validate([{ something: 'a' }]).$valid).toBe(true)
   })
 })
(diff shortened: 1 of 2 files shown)


haug1:
I've seen this now: https://github.com/vuelidate/vuelidate/pull/1050. Why is there not any progress on this clear and obvious error in the library for almost a year? Is there hope of getting this merged? If you're so concerned about the breaking change that the properties without validators return empty validation results now disappear, then there should be ways to work around that and still fix the bug. It would only take little more code to add the empty validation results for those properties. Personally, I think it's better to just remove it because it's a pointless extra operation, but maybe there's some implications or functionality to it that I'm not thinking about..

Update: I now understand that it may break some validation functions, primarily those made specifically for usage inside a `helpers.forEach`? Because they were maybe implemented not expecting undefined to get passed as a value before? But we are migrating from v0.7.7 and trying to replace the usage with this one and it breaks. The documentation says it's there specifically as a migration path and discourages usage and instead opt for nested components. So I think the `helpers.forEach` function has never been the replacement it was intended to be. 

TL;DR: If this is decided to not be merged, could it maybe be a desirable option to expose a new function with this change implemented as a true migration option from v0.x, in case anyone else may stumble down the same path?

